PR #1, opened 2020-04-13
Enable credential passing for client

gluap:
The client works great for me, I love how intuitive it is to use. Since my mqtt server requires authentication and I saw no builtin way to use it I added it this way to the client startup.

In paho one calls _client.set_username_pw after the client is instantiated but with the context manager setting up the context fails. Because one never enterst the context one never gets a chance to pass the credentials to the client.


frederikaalund:
Thanks for the pull request.

Looks good to me. :)

Is part of release v0.3.0 (available now).

PR #3, opened 2020-05-05
Pass received message information onward instead of only the payload

### LINKED ISSUE #2: Provide information about message instead of only the payload in unfiltered_messages
When using `unfiltered_messages` to receive all MQTT messages it would be useful if information about the message was provided instead of just the payload. For example, the `topic` field of the message at minimum would be useful.

This could be achieved by passing `msg` to `messages.put_nowait` in [`_cb_and_generator`](https://github.com/sbtinstruments/asyncio-mqtt/blob/master/asyncio_mqtt/client.py#L136) instead of `msg.payload`. The downside is that this exposes an instance of a class from mqtt.paho.

MBradbury:
When using `unfiltered_messages` to receive all MQTT messages it would be useful if information about the message was provided instead of just the payload. This PR changes the implementation to pass on the instance of MQTTMessage from paho instead of only the payload of the message.

Fixes #2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,12 @@ from asyncio_mqtt import Client
 async def log_filtered_messages(client, topic_filter):
     async with client.filtered_messages(topic_filter) as messages:
         async for message in messages:
-            print(f'[topic_filter="{topic_filter}"]: {message.decode()}')
+            print(f'[topic_filter="{topic_filter}"]: {message.payload.decode()}')
 
 async def log_unfiltered_messages(client):
     async with client.unfiltered_messages() as messages:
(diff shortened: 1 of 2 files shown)


frederikaalund:
Thanks for this. It's a nice API change. :)

I'll add a changelog to keep track of such changes.

frederikaalund:
This is part of release 0.4.0 (available now).

NJannasch:
Exactly what I was looking for!

Trance-Paradox:
How can I extract the topic and payload from unfiltered message?

PR #7, opened 2020-06-02
Enable python 3.6 and mutual TLS

pallas:


frederikaalund:
Thanks for taking interest in the API design and backwards compatibility with Python 3.6.

This is some nice additions and the implementation of them looks good to me.

Please keep the pull requests coming if you find other things, that you would like added or changed. 👍

pallas:
Thanks, will do!

PR #8, opened 2020-06-24
docs: update changelog

pallas:
s/gluap/pallas/


frederikaalund:
Thanks for that, and sorry for messing up the link in the first place. My bad. At least I got the username right in the text itself. :)

PR #12, opened 2020-08-01
fix: python 3.6 compat in _message_generator

pallas:
In `e645297 feat: improve reconnection support (yet again)`, python3.6 got
broken because of a call to `asyncio.create_task`, which only exists in
python 3.7.  The fix is to replace it with `self._loop.create_task`.

Signed-off-by: Derrick Lyndon Pallas <derrick@meter.com>

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -189,7 +189,7 @@ class Client:
                 # Wait until we either:
                 #  1. Receive a message
                 #  2. Disconnect from the broker
-                get = asyncio.create_task(messages.get())
+                get = self._loop.create_task(messages.get())
                 try:
                     done, _ = await asyncio.wait(
                         (get, self._disconnected),


pallas:
I swear I'm trying to move on to python 3.7!

frederikaalund:
Thanks for creating this pull request.

> In e645297 feat: improve reconnection support (yet again), python3.6 got
broken because of a call to asyncio.create_task, which only exists in
python 3.7

Ah, you're absolutely right. Sorry about that. I do try to maintain Python 3.6 compatibility but I must have missed that one.

Currently, there are no automated tests/builds and I only test on the latest stable Python version before release. An automated build/test system would definitely catch these issues but I don't have personal/commercial motivation to implement this. I'll be glad to accept any future pull requests that adds this.

> I swear I'm trying to move on to python 3.7!

Backward compatibility is a burden but an important one. :) Thanks for making these pull requests.

PR #17, opened 2020-08-25
add clean_session argument

madnadyka:


frederikaalund (changes requested):
Nice and clean implementation. :)

I have one request, though: Let's use `None` as the default value instead of `True`. This is also done in [paho.mqtt.Client](https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L512) and it triggers [special behaviour based on the other keyword arguments](https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L547).

madnadyka:
Yes, it will be better to use None! Thank you!

frederikaalund:
Looks good to me. Thanks for the pull request. :)

Let me know if you run into other issues.

PR #21, opened 2020-11-05
Added transport option to Client construnctor

opengs:
Small fix adds "transport" option to the constructor.
This allows to use "websocket" transport layer instead of default "tcp".


opengs:
Also it would be cool to have new release, because there were several minor fixes (Maybe v0.7.1?). It will be nice to have all this fixes accessible from the pip.

frederikaalund:
Thanks for the pull request. :) It looks good to me.

> Also it would be cool to have new release, because there were several minor fixes (Maybe v0.7.1?). It will be nice to have all those fixes accessible from the pip.

I agree. I'll make a new release in the coming days.

frederikaalund:
@opengs Version 0.8.0 is now up. :)

PR #36, opened 2021-02-15
Fix socket options for websocket transport

chmielowiec:
Fixing the following error:

```
  File /venv/lib/python3.8/site-packages/asyncio_mqtt/client.py", line 81, in connect
    self._client.socket().setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)
AttributeError: 'WebsocketWrapper' object has no attribute 'setsockopt'
```

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -79,7 +79,7 @@ class Client:
             await loop.run_in_executor(None, self._client.connect, self._hostname, self._port, 60)
             # paho.mqttClient.socket() return non-None after the call to connect.
             client_socket = self._client.socket()
-            if type(client_socket) is not mqtt.WebsocketWrapper:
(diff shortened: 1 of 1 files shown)


frederikaalund:
Hi Robert. Thanks for the pull request. :+1:  Let me have a look.

Nice find. I never tested this with a WebSocket-wrapped connection.

I like the fix in general. One comment though: Let's use `isinstance` instead. I think that's more idiomatic. What do you think?

chmielowiec:
Thanks for quick review! Changed according to your proposal, thanks :-) (still learning python)

frederikaalund:
Super. :+1: Let me know how the test goes. I'll merge it afterwards.

chmielowiec:
Tested, and of course it works perfectly, thanks for the tip!

PR #37, opened 2021-02-24
Add type hints

flyte:
This adds type hints to `.pyi` files to avoid changing/cluttering up the source files.

I've gone through the paho code to verify the types used, but not tested it beyond running `mypy --strict asyncio_mqtt`.


frederikaalund:
Hi Ellis. Thanks for your pull request. Let me have a look. :)

Great to see some typing. :+1: I wanted to add this at one point or another myself. As for the types themselves, I have no objections. It looks good.

I'm wondering if we should simply inline the types instead of using stub files. Minimum Python version is 3.7 so that isn't a barrier. Likewise, we can use the built-in `typing` module.

What do you think? I know that you mention "avoid changing/cluttering" as the main reason. Are there other reasons for this preference?

Again, thanks for this pull request.

flyte:
I actually prefer doing it inline, as you don't need to keep two files in sync. I can move them over when I get a moment. I might also add hints for the private methods too, since it's helpful for development on this project as well as for projects using this client.

frederikaalund:
> I actually prefer doing it inline, as you don't need to keep two files in sync. 

I agree. :)

> I can move them over when I get a moment. I might also add hints for the private methods too, since it's helpful for development on this project as well as for projects using this client.

I would really appreciate that, thank you. :+1:

frederikaalund:
I can merge this right away. Let me know if you prefer that, or, if you prefer to change it into inline typings first.

flyte:
I'll change into inline tomorrow hopefully, so don't bother merging just yet. Cheers

flyte:
With a large change like this, it's super helpful to use auto-formatting (I can't imagine working without it any more). It may be a separate issue to raise, but what do you feel about using https://github.com/psf/black to format the code?

I'll make a draft PR in a minute to show you what the changes would be when using black to format the codebase.

frederikaalund:
It's a great idea. :+1: See my reply in #43 for details.

I feel that we align on all these recent issues. Do you want to co-maintain asyncio-mqtt with me? This way, you can freely make changes/improvements without delay. It's still the early days (pre v1.0.0), so there's plenty of opportunity to put a personal touch on the project (give you a feeling of ownership). :)

flyte:
> I feel that we align on all these recent issues

Yeah, it's slightly uncanny!

> Do you want to co-maintain asyncio-mqtt with me?

Sure! Thanks :grin: . I was part-way through writing my own wrapper around paho-mqtt before I found this one. I thought https://github.com/mossblaser/aiomqtt was the only one, and I thought (hoped, ignorantly) that I might be able to make something that integrates paho more closely with the asyncio event loop. Then I saw that you had beaten me to it!

I can't promise to always be engaged with the project; I tend to dip in and out of things a bit, but I'd be happy to help push it along.

frederikaalund:
You're now officially an asyncio-mqtt maintainer. :tada: Welcome on board.

I had a real close look at aiomqtt (and the [other alternative](https://github.com/sbtinstruments/asyncio-mqtt#alternative-asyncio-based-mqtt-clients)) as well before I created asyncio-mqtt. Great minds think alike. :))

> I can't promise to always be engaged with the project; I tend to dip in and out of things a bit, but I'd be happy to help push it along.

No worries; I do the same. Put in the time that you can afford and no more.

frederikaalund:
You can merge this PR yourself now when you see fit. :+1:

flyte:
Cool, thanks!

I wonder how you feel about enabling the new GitHub Discussions feature? I've just done it on [one of my repos](https://github.com/flyte/pi-mqtt-gpio/discussions) and it seems like a nice way to discuss things without making a bunch of issues.

frederikaalund:
I never tried GitHub Discussions but let's give it a go. :) I enabled it.

flyte:
I've added inline type hints throughout, and split up the `_pending_calls` list into:

```python
self._pending_subscribes: Dict[int, "asyncio.Future[int]"] = {}
self._pending_unsubscribes: Dict[int, asyncio.Event] = {}
self._pending_publishes: Dict[int, asyncio.Event] = {}
```

so that we don't have Dicts with different types of Futures and Events in. The correct dict gets passed in to `_pending_call()` so that the Future/Event can be added and removed.

I added a property that returns a set of `mid`s that have pending calls, so that we're still able to get a count of them and check if the `mid`'s already got a pending call:

```python
    @property
    def _pending_calls(self) -> Set[int]:
        """
        Return a set of all message IDs with pending calls.
        """
        mids: Set[int] = set()
        mids.update(self._pending_subscribes.keys())
        mids.update(self._pending_unsubscribes.keys())
        mids.update(self._pending_publishes.keys())
        return mids
```

`_cb_and_generator()` was a bit tricky to follow through, but I think I managed to end up with the right types for everything!

I know I can merge myself, but I'll leave it for you to review beforehand, in case there's something you want to discuss.

asyncio_mqtt/client.py
```diff
@@ -264,9 +333,19 @@ def _pending_call(self, mid, value):
             #
             # However, if the callback doesn't get called (e.g., due to a
             # network error) we still need to remove the item from the dict.
-            self._pending_calls.pop(mid, None)
+            try:
```
frederikaalund, line 336: I personally prefer the shorter `pending_dict.pop(mid, None)` compared to the four lines of `try..except`. I guess that `try..except` is more explicit, though. It'll do. 👍

asyncio_mqtt/client.py
```diff
@@ -302,67 +387,91 @@ def _on_disconnect(self, client, userdata, rc, properties=None):
...
             self._disconnected.set_exception(MqttCodeError(rc, "Unexpected disconnect"))
 
-    def _on_subscribe(self, client, userdata, mid, granted_qos, properties=None):
+    def _on_subscribe(
+        self,
+        client: mqtt.Client,
+        userdata: Any,
+        mid: int,
+        granted_qos: int,
+        properties: mqtt.Properties = None,
```
frederikaalund, line 396: Let's use `Optional[...]` here as well. :)

asyncio_mqtt/client.py
```diff
@@ -302,67 +387,91 @@ def _on_disconnect(self, client, userdata, rc, properties=None):
...
+    ) -> None:
         self._loop.remove_writer(sock)
 
-    async def _misc_loop(self):
+    async def _misc_loop(self) -> None:
         while self._client.loop_misc() == mqtt.MQTT_ERR_SUCCESS:
             await asyncio.sleep(1)
 
-    async def __aenter__(self):
+    async def __aenter__(self) -> "Client":
```
frederikaalund, line 467: Alternatively, use `from __future__ import annotations`. Just check if that's Python 3.7 compatible beforehand. I can't remember. :)

asyncio_mqtt/client.py
```diff
@@ -2,49 +2,90 @@
...
 MQTT_LOGGER.setLevel(logging.WARNING)
 
 
+# TODO: This should be a (frozen) dataclass (from Python 3.7)
+# when we drop Python 3.6 support
+class Will:
+    def __init__(
+        self,
+        topic: str,
+        payload: PayloadType = None,
```
frederikaalund, line 48: Let's use `Optional[...]` here as well. 👍

asyncio_mqtt/client.py
```diff
@@ -2,49 +2,90 @@
...
+        qos: int = 0,
+        retain: bool = False,
+        properties: mqtt.Properties = None,
```
frederikaalund, line 51: Also `Optional[...]` for this guy.

asyncio_mqtt/client.py
```diff
@@ -2,49 +2,90 @@
...
-        logger=None,
-        client_id=None,
-        tls_context=None,
-        protocol=None,
-        will=None,
-        clean_session=None,
-        transport="tcp",
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        logger: logging.Logger = MQTT_LOGGER,
```
frederikaalund, line 68: Just note that this is a breaking change since the user can no longer explicitly use `logger=None`. I'm not opposed to the change as such. We just have to remember this due to semantic versioning.

asyncio_mqtt/client.py
```diff
@@ -2,49 +2,90 @@
...
+        client_id: Optional[str] = None,
+        tls_context: Optional[ssl.SSLContext] = None,
+        protocol: ProtocolType = mqtt.MQTTv311,
```
frederikaalund, line 71: See the comment for `logger`. Same thing applies.

asyncio_mqtt/client.py
```diff
@@ -78,16 +117,27 @@ def __init__(
...
         Note that paho-mqtt stores the client ID as `bytes` internally.
         We assume that the client ID is a UTF8-encoded string and decode
         it first.
         """
-        return self._client._client_id.decode()
+        return cast(bytes, self._client._client_id).decode()
 
-    async def connect(self, *, timeout=10):
+    @property
+    def _pending_calls(self) -> Set[int]:
```
frederikaalund, line 130: I'd personally prefer a generator for this. Something like:
```python
yield from self._pending_subscribes.keys()
yield from self._pending_unsubscribes.keys()
yield from self._pending_publishes.keys()
```

Accomplishes the same thing but:
 * Is shorter
 * Doesn't allocate memory for a set

In any case, a `Set` will do fine as well. 👍

frederikaalund (approved):
Looks good to me. 👍 I only have some minor comments here and there. Nothing of real importance. Have a look if you want to.

You can merge this PR in when you feel like it. :)

asyncio_mqtt/client.py
flyte: Right, so that's because the values in `pending_dict` can't be `None`, so the default causes a type error. Bit silly since we're not assigning it to anything, but there we go. I had to change it to this, although I agree using pop is nicer.

asyncio_mqtt/client.py
flyte: Oops, well spotted. Surprised that didn't give an error.

asyncio_mqtt/client.py
flyte: Yeah, I do tend to, but I've been writing 3.6+ code lately. What about this project is 3.7+ only? I ask because I've used it in my 3.6+ project :facepalm:

asyncio_mqtt/client.py
flyte: Oops again!

asyncio_mqtt/client.py
flyte: Argh :man_facepalming:

asyncio_mqtt/client.py
flyte: Ah, I did wonder why that was `None`. I'll revert that.

flyte: OK no probs, I'll revert that.

asyncio_mqtt/client.py
flyte: Will revert.

asyncio_mqtt/client.py
flyte: Ah yes, that's a nice solution.

frederikaalund:
Thanks for the huge effort on this one. 👍 👍 

Really nice to have types throughout the API (and internally). That'll make maintenance way easier going forward. :)

asyncio_mqtt/client.py
frederikaalund: I just checked to be sure: [`from __future__ import annotations` seems to be a Python 3.7 thing](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations). :)

> What about this project is 3.7+ only?

I think asyncio-mqtt (mostly) Python 3.6 compatible thanks to the efforts of pallas. I haven't kept track lately, though, so I might have broken 3.6 compatibility somewhere. 😅

Issue #39, opened 2021-02-24
Exception in on_socket_open: Non-thread-safe operation invoked on an event loop other than the current one


flyte:
When running with the `PYTHONASYNCIODEBUG=1`  environment variable set, using the following code:

```python
import asyncio

from asyncio_mqtt import Client


async def main():
    client = Client("test.mosquitto.org")

    print("Connecting...")
    await client.connect()
    print("Connected!")

    print("Sleeping...")
    await asyncio.sleep(5)
    print("Slept!")

    print("Disconnecting...")
    await client.disconnect()
    print("Disconnected!")


if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    try:
        loop.run_until_complete(main())
    finally:
        loop.close()
        loop.stop()
```

Connecting to an MQTT server raises the following exceptions:

```
$ PYTHONASYNCIODEBUG=1 PYTHONTRACEMALLOC=1 python3.8 test.py 
Connecting...
Caught exception in on_socket_open: Non-thread-safe operation invoked on an event loop other than the current one
Traceback (most recent call last):
  File "test.py", line 25, in <module>
    loop.run_until_complete(main())
  File "/usr/lib/python3.8/asyncio/base_events.py", line 608, in run_until_complete
    return future.result()
  File "test.py", line 10, in main
    await client.connect()
  File "/home/flyte/workspaces/asyncio-mqtt/ve/lib/python3.8/site-packages/asyncio_mqtt/client.py", line 79, in connect
    await loop.run_in_executor(None, self._client.connect, self._hostname, self._port, 60)
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/flyte/workspaces/asyncio-mqtt/ve/lib/python3.8/site-packages/paho/mqtt/client.py", line 941, in connect
    return self.reconnect()
  File "/home/flyte/workspaces/asyncio-mqtt/ve/lib/python3.8/site-packages/paho/mqtt/client.py", line 1117, in reconnect
    self._call_socket_open()
  File "/home/flyte/workspaces/asyncio-mqtt/ve/lib/python3.8/site-packages/paho/mqtt/client.py", line 2071, in _call_socket_open
    self.on_socket_open(self, self._userdata, self._sock)
[...]

PR #41, opened 2021-02-24
Check that _misc_task is not None before trying to cancel it.

flyte:
In #39 there's a second exception caused by `self._mqtt_task` being `None` when we try to cancel it. Since it's initialised as `None`, we shouldn't assume that it's not.

This is also something that would need to be fixed once type hints are added.


frederikaalund:
Nice find. :+1: Another win for type hints. :)

PR #43, opened 2021-02-25
Format code with black

flyte:
Just an example of what would change, as mentioned in #37 

I personally use a line length of 90, but it makes no difference to me to use a different one.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -312,7 +312,9 @@ class Client:
         try:
             self._pending_calls.pop(mid).set()
         except KeyError:
-            MQTT_LOGGER.error(f'Unexpected message ID "{mid}" in on_unsubscribe callback')
+            MQTT_LOGGER.error(
+                f'Unexpected message ID "{mid}" in on_unsubscribe callback'
+            )
 
     def _on_publish(self, client, userdata, mid):
         try:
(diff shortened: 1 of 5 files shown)


frederikaalund:
Good idea. :+1: 

I think the default line length in black is 88, right? I can't remember the rationale for that specific number. Just to be pragmatic, I vote that we stick to the default of 88. Less potential future conflicts.

I personally use black for other projects as well. asyncio-mqtt is really not the best showcase of my usual Python code style preferences. :)) Have a look at [this pyproject.toml](https://github.com/sbtinstruments/cyto/blob/master/pyproject.toml) for my preferred tools/defaults.

In a similar vein, we could also move to the poetry dependency manger, add isort, etc. I just never got around to "modernizing" asyncio-mqtt.

flyte:
Yep, I've redone it as 88 (it only changed one line!).

I also use poetry generally, although I haven't delved as deeply into using the pyproject.toml file as you have! I like that everything's contained there.

flyte:
Phew, it's still *just* under 400 lines of code :sweat_smile:

frederikaalund:
It used to be 300 lines of code. :sweat_smile: It keeps increasing as we add features and fix bugs. :) That's perfectly fine with me. The LoC count is just there to give an indication of: "This library is small and easy to improve/change/maintain. Jump right in!"

Issue #35, opened 2021-02-02
Broken pipe


gilbertsmink:
After some days of running I am receiving the "Broken pipe" error. This is currently not caught by asyncio-mqtt. Tbh this is too low level for me to understand. I expect it has something to do with an overflow of messages. It occurs at the same moment, I am not sure which is the cause and which is the effect, that the cpu, mem and swap are suddenly going crazy.

- Should this be something that my own script should catch and deal with?
- Should this be something that asyncio-mqtt should catch and deal with?
- or, is there another more preferred solution?

Trace:

```
evss_device_center.1.zaqg6zdkswua@evss24    | BrokenPipeError: [Errno 32] Broken pipe
evss_device_center.1.zaqg6zdkswua@evss24    | Exception in callback Client._on_socket_open.<locals>.cb() at /usr/local/lib/python3.8/site-packages/asyncio_mqtt/client.py:291
evss_device_center.1.zaqg6zdkswua@evss24    | handle: <Handle Client._on_socket_open.<locals>.cb() at /usr/local/lib/python3.8/site-packages/asyncio_mqtt/client.py:291>
evss_device_center.1.zaqg6zdkswua@evss24    | Traceback (most recent call last):
evss_device_center.1.zaqg6zdkswua@evss24    |   File "/usr/local/lib/python3.8/asyncio/events.py", line 81, in _run
evss_device_center.1.zaqg6zdkswua@evss24    |     self._context.run(self._callback, *self._args)
evss_device_center.1.zaqg6zdkswua@evss24    |   File "/usr/local/lib/python3.8/site-packages/asyncio_mqtt/client.py", line 292, in cb
evss_device_center.1.zaqg6zdkswua@evss24    |     client.loop_read()
evss_device_center.1.zaqg6zdkswua@evss24    |   File "/usr/local/lib/python3.8/site-packages/paho/mqtt/client.py", line 1572, in loop_read
evss_device_center.1.zaqg6zdkswua@evss24    |     rc = self._packet_read()
evss_device_center.1.zaqg6zdkswua@evss24    |   File "/usr/local/lib/python3.8/site-packages/paho/mqtt/client.py", line 2310, in _packet_read
evss_device_center.1.zaqg6zdkswua@evss24    |     rc = self._packet_handle()
evss_device_center.1.zaqg6zdkswua@evss24    |   File "/usr/local/lib/python3.8/site-packages/paho/mqtt/client.py", line 2936, in _packet_handle
[...]

PR #51, opened 2021-03-08
added exception handling in callback for loop_read and loop_write

gilbertsmink:
Proposed fix for issue #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ asyncio-mqtt combines the stability of the time-proven [paho-mqtt](https://githu
 * Fully type-hinted
 * Did we mention no more callbacks?
 
-The whole thing is less than [500 lines of code](https://github.com/sbtinstruments/asyncio-mqtt/blob/master/asyncio_mqtt/client.py).
+The whole thing is less than [600 lines of code](https://github.com/sbtinstruments/asyncio-mqtt/blob/master/asyncio_mqtt/client.py).
 
 ## Installation 📚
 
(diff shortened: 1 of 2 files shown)


frederikaalund:
Hi Gilbert. Thanks for making this pull request. :+1: 

I can see that there are some conflicts. It's probably because we added types to asyncio-mqtt recently. Can you try to rebase your changes on top of the latest commit on the master branch?

Otherwise, I think that this PR looks good. :) I'd like to see a comment at the `try..except` block that explains why we need it. Something like: "`loop_read` may raise an exception such as `BadPipe`. It's usually a sign that the underlying connection broke. If this happens, we disconnect right away."

gilbertsmink:
ok updated the pull request

frederikaalund:
Thanks for the effort, Gilbert. :+1: 

There are still some conflicts that prevent the merge. I see this in the GitHub UI:
image

Looks like you have several commits on this branch now. Some of them are unrelated to your fix. That's probably what caused the conflicts.

I think that we only need commit b4170fc873994ec129efc50f3512662def108836 of yours and not the rest.

---

I see three options here:
 1. You fix your pull request branch so that it only contains commit b4170fc873994ec129efc50f3512662def108836
 2. You close this pull request and create a new one that only contains commit b4170fc873994ec129efc50f3512662def108836
 3. I apply commit b4170fc873994ec129efc50f3512662def108836 manually on top of master. I'll give you full credit in the changelog of course.

Which option should we pursue? :)

gilbertsmink:
Ow Github is telling me there are no conflicts, but I hardly understand how rebase actually works. I will try again from skratch, because I have to climb this Git learning curve at some point, might as well start not

PR #54, opened 2021-03-16
Fix client socket defaults for websocket transport

chmielowiec:
After latest refactor websocket transport is broken again


frederikaalund:
Hi Robert, thanks for the pull request. Let me have a look. :)

Nice catch! :+1: That if statement is a mistake on my part. I simply forgot to put a `not` in front. Can we change it to the following:

```python
if not isinstance(client_socket, socket.socket):
```

Note the `not` in front of `isinstance`.

chmielowiec:
Changed according to your review.

PR #55, opened 2021-03-19
added exception handling in callback for loop_read and loop_write

gilbertsmink:
Proposed fix for issue #35


frederikaalund:
Thanks for the fix. :)

PR #56, opened 2021-03-27
Add ability to pass arguments to underlying client's connect method.

yawor:
The MQTTv5 protocol no longer uses `clean_session` setting of the client - it's prohibited by the underlying client to even set this argument when using MQTTv5 protocol. Instead, the MQTTv5 has a `clean_start` argument which has to be passed to the `connect` method of the client.
The asyncio-mqtt `Client` class doesn't provide any way to pass extra arguments to underlying client's `connect` method. This prevents using (amongst other) the clean_start functionality with MQTTv5 protocol.
This PR proposes a way to set arguments for the underlying paho client connect method, including the `clean_start` one, by providing a `ConnectArgs` class containing all the extra args and adds an optional `connect_args` argument to `Client` constructor.

An example usage:
```python
async def main():
    async with Client("localhost", protocol=MQTTv5, connect_args=ConnectArgs(clean_start=True)) as client:
        async with client.filtered_messages("foo/#") as messages:
            await client.subscribe("foo/#", options=SubscribeOptions(qos=1))
            async for message in messages:  # type: MQTTMessage
                print(message.topic, message.payload)
```

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -5,6 +5,7 @@ import socket
 import ssl
 from contextlib import contextmanager, suppress
 from enum import IntEnum
+from functools import partial
 from types import TracebackType
 from typing import (
     Any,
@@ -23,12 +24,14 @@ from typing import (
     cast,
 )
 
+
 try:
     from contextlib import asynccontextmanager
 except ImportError:
     from async_generator import asynccontextmanager  # type: ignore
 
 import paho.mqtt.client as mqtt  # type: ignore
+from paho.mqtt.properties import Properties
 
 from .error import MqttCodeError, MqttConnectError, MqttError
 from .types import PayloadType, T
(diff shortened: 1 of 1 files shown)


frederikaalund:
Hi Marcin. Thanks for creating this pull request. Let me have a look. :)

It's a good idea to expose the remaining `paho.Client.__init__` arguments. :+1: So far, we've done this via keyword arguments in `asyncio_mqtt.Client.__init__`. I see that you chose a slightly different approach with the `ConnectArgs` class as a layer of indirection. I suggest that we remove that layer of indirection and instead expose the remaining arguments directly. E.g.:

```python

class Client:
    def __init__(
        self,
        hostname: str,
        port: int = 1883,
        *,
        # All the existing `paho.Client` arguments
        username: Optional[str] = None,
        password: Optional[str] = None,
        logger: Optional[logging.Logger] = None,
        client_id: Optional[str] = None,
        tls_context: Optional[ssl.SSLContext] = None,
        protocol: Optional[ProtocolVersion] = None,
        will: Optional[Will] = None,
        clean_session: Optional[bool] = None,
        transport: str = "tcp",
        # The new arguments
        keepalive: Optional[int] = None,
        bind_address: Optional[str] = None,
        bind_port: Optional[int] = None,
        clean_start: Optional[bool] = None,
        properties: Optional[Properties] = None,
    ):
```

What do you think? :)

yawor:
That should be also ok. I wanted to somehow emphasize that these arguments are actually for connect in opposition to the Client instance itself.
But asyncio-mqtt API doesn't do the distinction anyway, as the client starts connection implicitly as a context manager, so maybe there's no need to complicate this. If you take this route then I think it'd be better to also copy default values for these args from the paho.
In the end, what's important is to be able to set these arguments, especially clean_start, as it can affect client logic a lot.

frederikaalund:
Thanks, I merged it to master. :)

PR #57, opened 2021-04-06
Fix python3.6 compatibility

fipwmaqzufheoxq92ebc:
The library currently uses methods like ```get_running_loop()```, which are only available in python>=3.7
This PR fixes that.

Fixes the downstream-issue https://github.com/flyte/mqtt-io/issues/192

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -162,7 +162,12 @@ class Client:
 
     async def connect(self, *, timeout: int = 10) -> None:
         try:
-            loop = asyncio.get_running_loop()
+            # get_running_loop is preferred, but only available in python>=3.7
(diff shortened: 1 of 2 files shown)


flyte:
Nice one @fipwmaqzufheoxq92ebc ! Looks good to me, but I'll hang on until @frederikaalund has a look :smiley:

frederikaalund:
Thanks for this pull request. It looks good to me. :+1: 

Do note that asyncio-mqtt officially targets Python 3.7. However, in the spirit of pragmatism:
 * We'll gladly merge in small compatibility fixes like this (thanks for the contribution :+1:)
 * We'll try not to break backwards compatibility (this is on a best-effort basis).

PR #59, opened 2021-05-04
handle rc ReasonCodes

### LINKED ISSUE #58: client._on_connect exception rc type error
It seems that https://github.com/sbtinstruments/asyncio-mqtt/blob/master/asyncio_mqtt/client.py#L387 may yield an error as `rc` is `paho.mqtt.reasoncodes.ReasonCodes` and not an integer?

    asyncio-mqtt==0.9.0
    paho-mqtt==1.5.1

jschlyter:
closes #58

### DIFF
--- a/asyncio_mqtt/error.py
+++ b/asyncio_mqtt/error.py
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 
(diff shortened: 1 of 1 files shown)


frederikaalund:
Looks good to me. Thanks. :+1:

PR #60, opened 2021-05-06
Improve cleanup when socket is invalid

wrobell:
Remove Asyncio loop reader for a socket, when its file number is valid as `socket.fileno` can return -1, see https://docs.python.org/3/library/socket.html#socket.socket.fileno.

This can happen because creation of socket can be delayed due to sleep  statements in Client._misc_loop method.

Also, cancel misc task only when it is not done.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -476,8 +476,11 @@ class Client:
     def _on_socket_close(
         self, client: mqtt.Client, userdata: Any, sock: socket.socket
     ) -> None:
-        self._loop.remove_reader(sock.fileno())
-        if self._misc_task is not None:
+
+        fileno = sock.fileno()
+        if fileno > -1:
+            self._loop.remove_reader(fileno)
+        if self._misc_task is not None and not self._misc_task.done():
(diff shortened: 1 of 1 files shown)


frederikaalund:
Thanks for this pull request. Let me have a look. :)

Did you run into an issue with this in practice? I'm wondering if we should also put a similar check [into the `_on_socket_open` callback](https://github.com/sbtinstruments/asyncio-mqtt/blob/6098ee05231ea345451b06f63260a3f93cd40199/asyncio_mqtt/client.py#L467). In turn, I wonder whether a check in `_on_socket_open` defeats the purpose of a check in `_on_socket_close`

Otherwise, it looks good to me. :+1: Thanks for your contribution to asyncio-mqtt.

wrobell:
Yes, I did run into this problem. When you are creating multiple new connections (I am creating 12), then some of them are delayed as sleep function is hit in `Client._misc_loop`. And when I kill my software at this exact time, then socket close fails with invalid descriptor error.

IMHO, you need checks in both open and close socket methods. Anything can happen at any time, so...

Issue #63, opened 2021-06-07
Publish Example, protocol Version as required field and logging


tropxy:
Hi,

I just started using your lib and after struggling a bit I managed to do what I wanted. I have a topic that I subscribed to and I published some json data and printed it back in the subscriber:

Subscriber code:
```python
async with Client("www.maqiatto.com", username="username", password="password", protocol=ProtocolVersion.V31) as client:
    async with client.filtered_messages("dummy/contactor") as messages:
        await client.subscribe("dummy/#")
        async for message in messages:
            print(message.topic)
            print(json.loads(message.payload))
```

Publisher code:
```python
async with Client("www.maqiatto.com", username="username", password="password", protocol=ProtocolVersion.V31) as client:
    message = {"state": 3}
    await client.publish("dummy/contactor", payload=json.dumps(message), qos=2, retain=False)
```

This works and I get the following printed out in the terminal:
```shell
dummy/contactor
{'state': 3}
```
So, looks good ;)

But I ran into some issues before I could make it work:
1.  There is no example or test for publish, so I had to dive into the details of your lib and also paho-mqtt and figure out what to do. Also for the Client connection just by inspecting the code one can check that Client class also accepts username, password and other properties necessary for the connection. So, I guess if an example directory with different possibilities and even some more detailed doc page would suffice to solve this.
2. First I tried to connect without specifying the Protocol used and it actually connects to the broker without an issue. The problem later is that I send a message to the broker, the process finishes and no error whatsoever is raised and I received no message in the subscriber side. So, I was quite puzzled about what was wrong and it was not until I defined the version that started working. Thus, I think the protocol version shouldnt be an optional field
3. This brings me to the point where I tried to import ProtocolVersion Enum class directly from the lib package, but it wouldnt work, because ProtocolVersion is not part of the __all__ list in the init.py. So, I had to import it specifying the client script:
```python
from asyncio_mqtt.client import ProtocolVersion
```
I guess it would be better to add the ProtocolVersion to the init.py

4. Last issue was receiving the JSON payload. I tried to decode the message directly when I received it:
```python
async for message in messages:
            print(json.loads(message))
```

But this wouldnt work, because message is of MQTT data type. This was not clear for me and I had to search around to understand that the message received can be stripped into two properties: "topic" and "payload" and by decoding the payload it worked. Maybe is due to my inexperience that I didnt know that, but it would be cool if somewhere this is explained or showed in an example.

[...]

PR #65, opened 2021-06-18
Publisher/Subscribe example also for JSON data and ProtocolVersion in the init

tropxy:
docs: Added a Publisher example, how to handle with the Client arguments and added an example of how to deal with JSON data; included ProtocolVersion in the __init__.py so that we can directly summon the Enum from the package import

This addresses #63

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 
 Write code like this:
 
+
 ##### Subscriber
 ```python
 async with Client("test.mosquitto.org") as client:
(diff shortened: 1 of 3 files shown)


frederikaalund:
Looks good! :+1: 

I have two requests:
  * Let's move the examples to an `/examples` folder to not clutter `README.md`. I like to keep the readme simple. :)
  * I like the change to `__init__.py`. :+1: It's unrelated to the docs, though, so can you move it into a separate commit?

Does it make sense?

tropxy:
Yup, it all makes sense. I initially added an examples folder, but then usually one creates python scripts to exemplify whatever is the used case you trying to exemplify and in this case I thought it is just easier to show in a Markdown file how it works and add comments there as necessary. Do you think it would make sense to create an Examples.md? Or you really prefer to have scripts?

frederikaalund:
Both `.py` and `.md` works for me. :+1: Whatever you prefer. This is the start of the documentation that we don't currently have. :))

tropxy:
> Both `.py` and `.md` works for me. 👍 Whatever you prefer. This is the start of the documentation that we don't currently have. :))

I hope it looks better now

frederikaalund:
Looks good to me. Thanks for your contribution to asyncio-mqtt. :) :+1:

tropxy:
My pleasure! I hope we can do like a proper doc page in the future ;)

PR #69, opened 2021-07-06
Added new parameter 'message_retry_set' to the Client.__init__. Set the timeout in seconds before a message with QoS>0 is retried. 20 seconds by default.

xydan83:


frederikaalund:
Thanks for the pull request. Let me have a look. :)

Nice and simple. :+1: Thanks for your contribution to asyncio-mqtt.

Issue #70, opened 2021-07-08
Very slow publishing data more 100 MB.


xydan83:
Hello again! I noticed one thing, when I try to transfer a file larger than 100MB, it takes a very long time when publishing, it can take 3-5 minutes. If I use TLS, then 2-3 times longer. Moreover, the subscriber receives this file instantly. Any idea why this might be the case? Could it be the small size of the TCP allocated buffer?

![image](https://user-images.githubusercontent.com/86729310/124878832-3713ba80-dfd5-11eb-9334-38d7a491ffd2.png)

P.s. I`m using Ubuntu 20.04.1

frederikaalund:
Hi again. :) Thanks for raising this issue. Let me have a look.

I've never tested with packets of that size to be honest. The specification let's the messages be as large as 256 MiB but people usually send much smaller messages (well under 1 MiB).

In any case, let's rule out some potential bottlenecks first.

 1. What kind of transfer speeds do you get with alternative MQTT clients on the same network?
 2. Is the server and client on the same machine? (that would be good for network traffic)
 3. Is the machine a recent desktop/laptop computer? (and not, say, a Raspberry Pi or something)

I'm particularly interested in (1). It'll give us a rough estimate of what to expect in therms of transfer speeds of large messages.

> Any idea why this might be the case? Could it be the small size of the TCP allocated buffer?

That's a good guess. :+1: As a quick test, you can try to increase this to `20480`. Let me know what you figure out. :)

frederikaalund:
By the way, the only reason that `SNDBUF` is `2048` is because I got that setting from the [`loop_asyncio.py` example in paho.mqtt](https://github.com/eclipse/paho.mqtt.python/blob/0cc8581aa11dbea31364619c1bec6f920dbfec3e/examples/loop_asyncio.py#L89). I don't know why they change it from the default.

frederikaalund:
The `SNDBUF` default value on my system is `212992`! That's 104 times bigger than `2048`. Maybe we shouldn't change it from the default.

xydan83:
I tried increasing the buffer and it really increased the transfer rate. Now my file with TLS enabled was transferred in less than 3 minutes. Previously, I generally went into a timeout.

![image](https://user-images.githubusercontent.com/86729310/124911527-84a01f80-dff5-11eb-929b-81704b730d93.png)

Can we move the buffer setting to the __init__?

And, my answers:
[...]

PR #71, opened 2021-07-12
Added a new parameter for configuring socket connection.

xydan83:
From https://github.com/sbtinstruments/asyncio-mqtt/issues/70 issue.

        # By default, in order to optimize the transmission of small data packets,
        # the send buffer size is set to 2048 bytes. To transfer large data packets (>50MB),
        # it is recommended to increase the buffer value.
        # It is also possible to set all possible socket settings.
        # See https://docs.python.org/3/library/socket.html#socket.socket.setsockopt for more information.
        # If the socket_options is None, then the system defaults are used.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -20,7 +20,7 @@ from typing import (
     Tuple,
     Type,
     Union,
-    cast,
+    cast, Iterable,
 )
 
 
@@ -67,6 +67,13 @@ class Will:
         self.properties = properties
 
 
+# See the overloads of `socket.setsockopt` for details.
+SocketOption = Union[
+    Tuple[int, int, Union[int, bytes]],
+    Tuple[int, int, None, int],
+]
+
+
 class Client:
     def __init__(
         self,
@@ -88,13 +95,7 @@ class Client:
         clean_start: bool = mqtt.MQTT_CLEAN_START_FIRST_ONLY,
         properties: Optional[Properties] = None,
         message_retry_set: int = 20,
-        # By default, in order to optimize the transmission of small data packets,
-        # the send buffer size is set to 2048 bytes. To transfer large data packets (>50MB),
-        # it is recommended to increase the buffer value.
-        # It is also possible to set all possible socket settings.
-        # See https://docs.python.org/3/library/socket.html#socket.socket.setsockopt for more information.
-        # If the socket_options is None, then the system defaults are used.
-        socket_options: Optional[Tuple[int, int, Union[int, bytes]]] = (socket.SOL_SOCKET, socket.SO_SNDBUF, 2048),
+        socket_options: Optional[Iterable[SocketOption]] = (),
     ):
         self._hostname = hostname
         self._port = port
@@ -150,7 +151,7 @@ class Client:
             )
 
         self._client.message_retry_set(message_retry_set)
-        self._socket_options = socket_options
+        self._socket_options: Tuple[SocketOption] = tuple(socket_options)
 
     @property
     def id(self) -> str:
@@ -187,8 +188,7 @@ class Client:
                 self._clean_start, self._properties
             )
             client_socket = self._client.socket()
-            if self._socket_options is not None:
-                _set_client_socket_defaults(client_socket, self._socket_options)
+            _set_client_socket_defaults(client_socket, self._socket_options)
         # paho.mqtt.Client.connect may raise one of several exceptions.
         # We convert all of them to the common MqttError for user convenience.
         # See: https://github.com/eclipse/paho.mqtt.python/blob/v1.5.0/src/paho/mqtt/client.py#L1770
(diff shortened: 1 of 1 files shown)


asyncio_mqtt/client.py
```diff
@@ -87,6 +87,14 @@ def __init__(
...
         clean_start: bool = mqtt.MQTT_CLEAN_START_FIRST_ONLY,
         properties: Optional[Properties] = None,
+        message_retry_set: int = 20,
+        # By default, in order to optimize the transmission of small data packets,
+        # the send buffer size is set to 2048 bytes. To transfer large data packets (>50MB),
+        # it is recommended to increase the buffer value.
+        # It is also possible to set all possible socket settings.
+        # See https://docs.python.org/3/library/socket.html#socket.socket.setsockopt for more information.
+        # If the socket_options is None, then the system defaults are used.
+        socket_options: Optional[Tuple[int, int, Union[int, bytes]]] = (socket.SOL_SOCKET, socket.SO_SNDBUF, 2048),
```
frederikaalund, line 97: Good job with the initial typing here. :+1: I really like the details in the comment as well

I suggest that we drop the default `SO_SNDBUF` setting. Move it to the comment, though, as a nice example.

I also suggest that we expand on this, so that people can set multiple options using any of [the three overloads of setsockopt](https://docs.python.org/3/library/socket.html#socket.socket.setsockopt). First, we define:
```python
# See the overloads of `socket.setsockopt` for details.
SocketOption = Union[
    Tuple[int, int, Union[int, bytes]],
    Tuple[int, int, None, int],
]
```

Second, we accept an iterable of socket options in the `Client.__init__`.
```python
socket_options: Iterable[SocketOption],
```

asyncio_mqtt/client.py
```diff
@@ -539,7 +551,8 @@ async def __aexit__(
 _PahoSocket = Union[socket.socket, mqtt.WebsocketWrapper]
 
 
-def _set_client_socket_defaults(client_socket: Optional[_PahoSocket]) -> None:
+def _set_client_socket_defaults(client_socket: Optional[_PahoSocket],
+                                socket_param: Optional[Tuple[int, int, Union[int, bytes]]]) -> None:
```
frederikaalund, line 555: Now we can change this as well:
```python
socket_options: Iterable[SocketOption]
```

asyncio_mqtt/client.py
```diff
@@ -141,6 +149,9 @@ def __init__(
                 will.topic, will.payload, will.qos, will.retain, will.properties
             )
 
+        self._client.message_retry_set(message_retry_set)
+        self._socket_options = socket_options
```
frederikaalund, line 153: Change to `self._socket_options: Tuple[SocketOption] = tuple(socket_options)`

asyncio_mqtt/client.py
```diff
@@ -176,7 +187,8 @@ async def connect(self, *, timeout: int = 10) -> None:
                 self._clean_start, self._properties
             )
             client_socket = self._client.socket()
-            _set_client_socket_defaults(client_socket)
+            if self._socket_options is not None:
+                _set_client_socket_defaults(client_socket, self._socket_options)
```
frederikaalund, line 191: We don't need the check anymore. :) Just do:
```python
set_client_socket_defaults(client_socket, self._socket_options)
```

asyncio_mqtt/client.py
```diff
@@ -550,4 +563,4 @@ def _set_client_socket_defaults(client_socket: Optional[_PahoSocket]) -> None:
         return
     # At this point, we know that we got an actual socket. We change
     # some of the default options.
-    client_socket.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)
+    client_socket.setsockopt(*socket_param)
```
frederikaalund, line 566: Lastly, we apply all the socket options:
```
for socket_option in socket_options:
    client_socket.setsockopt(*socket_option)
```

This way, the user can define multiple socket options. :+1:

frederikaalund (changes requested):
Thanks for the pull request. This is a really cool feature and a nice addition to asyncio-mqtt. :+1: 

I've reviewed the changes and I'd like some minor changes here and there. Let me know what you think. :)

asyncio_mqtt/client.py
xydan83: Yes, is good idea!

asyncio_mqtt/client.py
xydan83: Cool! I didn’t think of this solution.

xydan83:
Thank you! Great changes. Thus, the setting will be more flexible.

frederikaalund:
Glad you like it. :) Let me know when you push the new changes. :+1: 

For now, there seems to be some merge conflicts. I recommend that you:

 * Rebase on top of master `git pull --rebase -i` (the `-i` is for interactive mode)
 * Force push your changes `git push --force`

That should also get rid of the "Added new parameter 'message_retry_set' to the Client._" commit that we already merged to master. :)

asyncio_mqtt/client.py
xydan83: I found it worth adding one more thing so that there would be no compatibility issues with the new library version when upgrading.
If we do this: `socket_options: Iterable[SocketOption],` then we get an error
**TypeError: __init__() missing 1 required keyword-only argument: 'socket_options'**, because we have no input parameter when constructing an object.

Maybe add a default value?
For example: `socket_options: Optional[Iterable[SocketOption]] = ()`

xydan83:
I applied the changes, check please:)

frederikaalund:
I merged your pull request. Thanks for you contribution to asyncio-mqtt. :) :+1:

xydan83:
> I merged your pull request. Thanks for you contribution to asyncio-mqtt. :)

Thanks for a great project! ))

Issue #73, opened 2021-08-16
Consider converting ReasonCodes into readable strings in MqttCodeError


CptSpaceToaster:
Hi, new to the project, and loving where it's going.

I thought about opening a PR to start... but maybe some context first might make more sense. When debugging my app, I ran into a situation where a timeout would cause paho to stop sending all messages. I was catching and logging the errors, but missed the subtle difference between the `MqttError` and `MqttCodeError` types behind the scenes.

For reasons still somewhat unknown, the reason code was being set to an integer (not a `mqtt.ReasonCode`) and the subtle `[Code: 4]` added to the `__str__()` had slipped by. Once I took a deeper look at what was happening, I saw the extra information, as well as a helper method in paho's source code: https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L183

The simple complaint here, is that `[Code: {}]` is a small addition to an error message, easy to miss, and a not-very-idiomatic error number, when it could instead be an english error message in instances where the `rc` value is actually an integer.

using Paho's `error_message` handler or using their builtin methods, we could change the `__str__()` method in `MqttCodeError` to look something a little like this:

```python
class MqttCodeError(MqttError):
    def __init__(self, rc: Union[int, mqtt.ReasonCodes], *args: Any):
        super().__init__(*args)
        self.rc = rc

    def __str__(self) -> str:
        if isinstance(self.rc, mqtt.ReasonCodes):
            return f"[code:{self.rc.value}] {str(self.rc)}"
        else:
            return f"[code:{self.rc}] {mqtt.error_string(self.rc)} {super().__str__()}"
```

But for such a simple change... I figured it'd be better to ask if there was a reason it was setup this way, or if I'm touching something that hasn't been considered in a while. Maybe my versions are worth considering? `paho-mqtt-1.5.1` and `asyncio_mqtt-0.9.1`

If this sounds like a useful change, I'd be more than willing to open up a simple PR... otherwise I'd be interested in learning why the integer codes are preferred to Paho's internal error messages.

frederikaalund:
Hi CptSpaceToaster, thanks for opening this issue. Let me have a look. :)

That's a really good idea. I'd love to have that as part of asyncio-mqtt. Feel free to open a pull request. :+1: 

Sorry about the lackluster `MqttCodeError`. It's certainly not a usability master piece. :sweat_smile:

CptSpaceToaster:
[...]

PR #74, opened 2021-08-17
Convert reason codes into english error messages

CptSpaceToaster:
Per https://github.com/sbtinstruments/asyncio-mqtt/issues/73


frederikaalund:
Looks good, thanks. Good idea to drop the `super().__str__()` thing. There's no added value in that. :+1: 

Thank you for your contribution to asyncio-mqtt.

PR #86, opened 2021-11-05
Fix typing, mostly related to async_generator

laundmo:
While running the "advanced" example from the readme with pyright, i noticed it was alerting me to some typing issues, relating to asynccontextmanager and `Client.__aexit__`.

<details>
  <summary>Pyright errors</summary>

```
Argument of type "Client" cannot be assigned to parameter "cm" of type "AsyncContextManager[_T@enter_async_context]" in function "enter_async_context"
  "Client" is incompatible with protocol "AsyncContextManager[_T@enter_async_context]"
    "__aexit__" is an incompatible type
      Type "(exc_type: Type[Exception], exc: Exception, tb: TracebackType) -> Coroutine[Any, Any, None]" cannot be assigned to type "(__exc_type: Type[BaseException] | None, __exc_value: BaseException | None, __traceback: TracebackType | None) -> Awaitable[bool | None]"
        Parameter 1: type "Type[BaseException] | None" cannot be assigned to type "Type[Exception]"
          Type "Type[BaseException] | None" cannot be assigned to type "Type[Exception]"
        Parameter 2: type "BaseException | None" cannot be assigned to type "Exception"
          Type "BaseException | None" cannot be assigned to type "Exception"
        Parameter 3: type "TracebackType | None" cannot be assigned to type "TracebackType"
```
```
Argument of type "AsyncIterator[AsyncGenerator[MQTTMessage, None]]" cannot be assigned to parameter "cm" of type "AsyncContextManager[_T@enter_async_context]" in function "enter_async_context"
  "AsyncIterator[AsyncGenerator[MQTTMessage, None]]" is incompatible with protocol "AsyncContextManager[_T@enter_async_context]"
    "__aenter__" is not present
    "__aexit__" is not present
```
```
Argument of type "AsyncIterator[AsyncGenerator[MQTTMessage, None]]" cannot be assigned to parameter "cm" of type "AsyncContextManager[_T@enter_async_context]" in function "enter_async_context"
  "AsyncIterator[AsyncGenerator[MQTTMessage, None]]" is incompatible with protocol "AsyncContextManager[_T@enter_async_context]"
    "__aenter__" is not present
    "__aexit__" is not present
```
</details>

The issue with `Client.__aexit__` was simple, the arguments were missing the `Optional[]` around their type hint, and also need to use `BaseException` instead of `Exception`.

After a LOT of time trying to figure out what the issue wth asynccontextmanager was, i noticed that the compatibility import `async_generator` was missing type information, which caused pyright to fail at inferring the `AsyncContextManager` type for `Client.filtered_messages` and `Client.unfiltered_messages`. 

I solved this by redefining the asynccontextmanager imported from `async_generator` with the proper typehints, and ignoring the type issues on those lines. This could potentially be moved to a type stub, though im not very comfortable working with them yet.

After these changes there are no more issues running pyright on the advanced readme example.  
MyPy still throws a few errors relating to SocketOption, but this is behaviour that has not changed.
I have tested running the advanced readme example on python 3.6 and with modifications to the example due to missing features it runs fine.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -21,13 +21,20 @@ from typing import (
     Type,
     Union,
     cast, Iterable,
+    AsyncContextManager,
+    TypeVar,
 )
+from typing_extensions import ParamSpec
 
 
 try:
     from contextlib import asynccontextmanager
 except ImportError:
-    from async_generator import asynccontextmanager  # type: ignore
+    from async_generator import asynccontextmanager as _asynccontextmanager # type: ignore
+    _P = ParamSpec('_P')
+    _T = TypeVar('_T')
+    def asynccontextmanager(func: Callable[_P, AsyncIterator[_T]]) -> Callable[_P, AsyncContextManager[_T]]: # type: ignore
+        return _asynccontextmanager(func)
 
 import paho.mqtt.client as mqtt  # type: ignore
 from paho.mqtt.properties import Properties
@@ -92,7 +99,7 @@ class Client:
         keepalive: int = 60,
         bind_address: str = "",
         bind_port: int = 0,
-        clean_start: bool = mqtt.MQTT_CLEAN_START_FIRST_ONLY,
+        clean_start: int = mqtt.MQTT_CLEAN_START_FIRST_ONLY,
         properties: Optional[Properties] = None,
         message_retry_set: int = 20,
         socket_options: Optional[Iterable[SocketOption]] = (),
@@ -527,7 +534,7 @@ class Client:
         return self
 
     async def __aexit__(
-        self, exc_type: Type[Exception], exc: Exception, tb: TracebackType
+        self, exc_type: Optional[Type[BaseException]], exc: Optional[BaseException], tb: Optional[TracebackType]
     ) -> None:
         """Disconnect from the broker."""
         # Early out if already disconnected...
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -24,13 +24,14 @@ from typing import (
     AsyncContextManager,
     TypeVar,
 )
-from typing_extensions import ParamSpec
+
 
 
 try:
     from contextlib import asynccontextmanager
 except ImportError:
     from async_generator import asynccontextmanager as _asynccontextmanager # type: ignore
+    from typing_extensions import ParamSpec
     _P = ParamSpec('_P')
     _T = TypeVar('_T')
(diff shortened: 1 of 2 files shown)


asyncio_mqtt/client.py
```diff
@@ -21,13 +21,21 @@
...
 try:
     from contextlib import asynccontextmanager
 except ImportError:
-    from async_generator import asynccontextmanager  # type: ignore
+    from async_generator import asynccontextmanager as _asynccontextmanager # type: ignore
+    from typing_extensions import ParamSpec
+    _P = ParamSpec('_P')
+    _T = TypeVar('_T')
+    def asynccontextmanager(func: Callable[_P, AsyncIterator[_T]]) -> Callable[_P, AsyncContextManager[_T]]: # type: ignore
+        return _asynccontextmanager(func)
```
sdwilsh, line 38: https://github.com/sbtinstruments/asyncio-mqtt/pull/87 has a less-verbose fix for this for python3.7+

I suspect some combination of the two is desired for support for 3.6 with type information.

laundmo: i did test this with 3.6, though not with a 3.6 pyright (I don't think it would make a difference)

this basically just copies the exact signature that the contextlib version has, so theres no need for pyright to understand sys.version, I don't think.

maybe it would still be better to use sys.version instead of the try-except though.

frederikaalund:
Hi @laundmo and @sdwilsh. Thanks for opening these pull requests. I'll tackle both #86 and #87 in here since they cover the same underlying issue.

Good catch on the type errors in general. :+1:  I'm happy to merge this as-is.

I guess the only open question is how to deal with the untyped `async_generator.asynccontextmanager` in python 3.6. Again, I'll happily accept either of the approaches that you propose. Do you guys agree on a specific approach or do you need a BDFL decision? :)

laundmo:
@frederikaalund why not both? replace the `try/except` with the version `if`, and leave my type overwrite in that. That way, 3.6 has correct typing and pyright can still understand it won't ever get to use `async_generator.asynccontextmanager` on versions > 3.6

frederikaalund:
You're right, we need both. :+1: I'll merge this as soon as that last change lands in this PR.

frederikaalund:
I'll credit you both in `CHANGELOG.md`, by the way. :)

laundmo:
@frederikaalund i don't have any more changes planned. Feel free to merge, or do you want me to merge the other PR into this one first?

PR #87, opened 2021-11-07
Pyright reports false positive error with `asynccontextmanager`

sdwilsh:
Pyright gets confused and will give you error messages like this:
```
error: Object of type "AsyncIterator[AsyncGenerator[...]]" cannot be used with "with" because it does not implement __aenter__ (reportGeneralTypeIssues)
error: Object of type "AsyncIterator[AsyncGenerator[...]]" cannot be used with "with" because it does not implement __aexit__ (reportGeneralTypeIssues)
```
For code like this:
```python
    async with client.filtered_messages("floors/+/humidity") as messages:
        await client.subscribe("floors/#")
            print(message.payload.decode())
```

This change statically checks against `sys.version_info` for the right version, which Pyright understands.

For additional discussion, see https://github.com/encode/httpx/discussions/1829.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -3,6 +3,7 @@ import asyncio
 import logging
 import socket
 import ssl
+import sys
 from contextlib import contextmanager, suppress
 from enum import IntEnum
 from types import TracebackType
@@ -24,9 +25,11 @@ from typing import (
 )
 
 
-try:
+
+# asynccontextmanager exists in contextlib in python3.7 and later.
+if sys.version_info >= (3, 7)
     from contextlib import asynccontextmanager
-except ImportError:
+else:
     from async_generator import asynccontextmanager  # type: ignore
 
 import paho.mqtt.client as mqtt  # type: ignore


sdwilsh:
Whoops...I had checked for issues on this, but failed to check PRs.  It looks like https://github.com/sbtinstruments/asyncio-mqtt/pull/86 by @laundmo is an alternative fix for this as well.

frederikaalund:
Thanks for opening this pull request. There is a lot of overlap with #86, yes, so let's take the discussion in there. :+1: Seems to me that @laundmo is open to your approach. Let's discuss in #86. :)

frederikaalund:
After some discussion in #86, I chose to merge #86 first. That puts this PR (#87) in conflict.

I'd like to merge it in, though. I think it's trivial to make it compatible with the current master. Can you please update this PR accordingly? Thanks! :)

frederikaalund:
Looks good to me. :) Thank you for your contribution to asyncio-mqtt. 👍

sdwilsh:
@frederikaalund, any thoughts on when you might be cutting a new release?

frederikaalund:
@sdwilsh, I'll make a new release first thing on Monday. :) I just need to update the changelog accordingly. Should take 10 mins.

frederikaalund:
I just released v0.11.1. Better late than never. :)

PR #93, opened 2021-12-01
fixes #92

### LINKED ISSUE #92: The advanced example is possibly buggy
Thank you for asyncio-mqtt; I've used it many of my home IoT applications.
I noticed one (pedantic) thing the the advance example that you might consider changing...
The function cancel_task contains the code:

        task.cancel()
        try:
            await task
        except asyncio.CancelledError:
            pass

Unfortunately, it is possible that the CancelledError exception arrives between `task.cancel` call and the try block.  It is better practice to do:

        try:
            task.cancel()
            await task
        except asyncio.CancelledError:
            pass

steverpalmer:
As discussed...

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ async def cancel_tasks(tasks):
     for task in tasks:
         if task.done():
             continue
-        task.cancel()
         try:
+            task.cancel()
             await task
         except asyncio.CancelledError:
             pass


frederikaalund:
Looks good to me. :) Thank you for your contribution to asyncio-mqtt. 👍

Issue #29, opened 2020-12-30
pending publish calls


vitalerter:
Hi,
what is this warning means?
I am publishing with qos1
at:
`/usr/local/lib/python3.7/site-packages/asyncio_mqtt/client.py`

`There are 13 pending publish calls.`
`There are 12 pending publish calls.`

frederikaalund:
Hi vitalerter. Thanks for submitting this issue. Let me have a look. :+1: 

This "pending publish calls" warning appears when there are more than ten messages in the "publish queue". It is basically a sign of congestion. In other words: You publish messages faster than the system can process them.

If the number of "pending publish calls":
 * increases over time, you must take action. The process will run out of memory at some point in time.
 * decreases over time, you don't have to worry.

There is no back pressure in asyncio-mqtt as it is now. It would be pretty straight-forward to add a global back pressure system for all the possible "pending calls". Simply fix the `max_size` parameter of the pending calls queue. I just never got around to do that. Do you want to have a go at it?

~Frederik

vitalerter:
it is occurs when publishing with qos>0

```
async def mqtt_publish(self, topic, msg, broker):
        try:
            res = await self.clients.get(broker).publish(topic, msg, qos=1)
            _logging.debug("Mqtt publishing", extra={
                "data": {"result": res, "message": msg, "topic": topic}})
        except MqttError as identifier:
            _logging.error("Mqtt publishing Failed", extra={
                "data": {"exception": identifier, "message": msg, "topic": topic}})
```

frederikaalund:
> it is occurs when publishing with qos>0

That does make sense, since `qos > 0` requires more communication (and hence more time to complete). :)

Again, this warning is a sign of congestion. How many messages are you publishing and how fast are you publishing them?

vitalerter:
Hi.
I  increased the `self._pending_calls_threshold = 50` this is not showing me the message.

frederikaalund:
> I increased the self._pending_calls_threshold = 50 this is not showing me the message.

Question is then whether:
 * You'll still see the warning but just at a much later time (since the threshold is much larger now)
 * The threshold is now large enough to account for time-limited congestion.

In any case, I'm glad that it worked out for you. I don't want to increase the default threshold just yet as I still believe the current threshold fits the common use case. If we get additional reports about these warnings, then I'll do one or more of the following:
 * Reconsider the default
 * Provide an API to change the default threshold
[...]

PR #101, opened 2022-01-17
Add backpressure mechanism to limit the number of concurrent outgoing calls

bachya:
This PR applies a backpressure mechanism so that extremely chatty clients don't add congestion to the server. The mechanism is controlled with a `max_concurrent_outgoing_calls` parameter that can be applied to the client during instantiation:

* A `max_concurrent_outgoing_calls` value of `None` means that no back pressure is applied
* A `max_concurrent_outgoing_calls` value with a positive `int` indicates that no more than `max_concurrent_outgoing_calls` outgoing calls can happen at any given time.

Fixes https://github.com/sbtinstruments/asyncio-mqtt/issues/29

@frederikaalund Some comments for your review and consideration:

* As discussed, the fact that `max_concurrent_outgoing_calls` can be `None` (and is so by default) should retain backward compatibility.
* I didn't see any other spots in the README where specific functionality is shown via example; let me know if you think one is warranted here.
* I didn't apply any linting/styling since I couldn't determine whether a standard was already used. I did endeavor to follow the existing standard for type hinting.
* I didn't add any tests for this since the library as a whole doesn't seem to carry many; let me know if you would prefer I add some.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -105,6 +105,7 @@ class Client:
         properties: Optional[Properties] = None,
         message_retry_set: int = 20,
         socket_options: Optional[Iterable[SocketOption]] = (),
+        max_concurrent_outgoing_calls: Optional[int] = None,
     ):
         self._hostname = hostname
         self._port = port
@@ -122,6 +123,14 @@ class Client:
         self._pending_publishes: Dict[int, asyncio.Event] = {}
         self._pending_calls_threshold: int = 10
         self._misc_task: Optional["asyncio.Task[None]"] = None
+
+        self._outgoing_calls_sem: Optional[asyncio.Semaphore]
+        if max_concurrent_outgoing_calls:
+            self._outgoing_calls_sem = asyncio.Semaphore(
+                max_concurrent_outgoing_calls
+            )
+        else:
+            self._outgoing_calls_sem = None
 
         if protocol is None:
             protocol = ProtocolVersion.V311
@@ -218,6 +227,18 @@ class Client:
         if not self._disconnected.done():
             self._disconnected.set_result(None)
 
+    def _outgoing_call(self, func: Callable) -> Callable[..., Awaitable]:
+        async def decorated(*args: Any, timeout: int = 10, **kwargs: Any) -> None:
+            if not self._outgoing_calls_sem:
+                func(*args, timeout=timeout, **kwargs)
+                return
+
+            async with self._outgoing_calls_sem:
+                func(*args, timeout=timeout, **kwargs)
+
+        return decorated
+
+    @_outgoing_call
     async def subscribe(self, *args: Any, timeout: int = 10, **kwargs: Any) -> int:
         result, mid = self._client.subscribe(*args, **kwargs)
         # Early out on error
@@ -229,6 +250,7 @@ class Client:
             # Wait for cb_result
             return await self._wait_for(cb_result, timeout=timeout)
 
+    @_outgoing_call
     async def unsubscribe(self, *args: Any, timeout: int = 10) -> None:
         result, mid = self._client.unsubscribe(*args)
         # Early out on error
@@ -240,6 +262,7 @@ class Client:
             # Wait for confirmation
             await self._wait_for(confirmation.wait(), timeout=timeout)
 
+    @_outgoing_call
     async def publish(self, *args: Any, timeout: int = 10, **kwargs: Any) -> None:
         info = self._client.publish(*args, **kwargs)  # [2]
         # Early out on error
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -233,7 +233,7 @@ class Client:
     # 3.10). See: https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext
     def _outgoing_call(self, func: Callable) -> Callable[..., Awaitable]:
         @functools.wraps(func)
-        async def decorated(*args: Any, **kwargs: Any) -> None:
+        async def decorated(*args: Any, **kwargs: Any) -> Union[None, int]:
             if not self._outgoing_calls_sem:
                 return await func(*args, **kwargs)
 
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -64,8 +64,6 @@ class ProtocolVersion(IntEnum):
 # TODO: This should be a (frozen) dataclass (from Python 3.7)
 # when we drop Python 3.6 support
 class Will:
-
-    OutgoingType = TypeVar("OutgoingType", None, int)
     def __init__(
         self,
         topic: str,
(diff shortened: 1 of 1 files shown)


asyncio_mqtt/client.py
```diff
@@ -123,6 +124,14 @@ def __init__(
         self._pending_calls_threshold: int = 10
         self._misc_task: Optional["asyncio.Task[None]"] = None
 
+        self._outgoing_calls_sem: Optional[asyncio.Semaphore]
+        if max_concurrent_outgoing_calls:
```
frederikaalund, line 128: Let's use `if max_concurrent_outgoing_calls is not None` instead. Otherwise, the value `0` also takes us to the `else` branch.

asyncio_mqtt/client.py
```diff
@@ -218,6 +227,18 @@ async def force_disconnect(self) -> None:
         if not self._disconnected.done():
             self._disconnected.set_result(None)
 
+    def _outgoing_call(self, func: Callable) -> Callable[..., Awaitable]:
+        async def decorated(*args: Any, timeout: int = 10, **kwargs: Any) -> None:
```
frederikaalund, line 231: Why is `timeout` mentioned explicitly here? Let's just omit it (let it go into the `kwargs`).

asyncio_mqtt/client.py
```diff
@@ -218,6 +227,18 @@ async def force_disconnect(self) -> None:
         if not self._disconnected.done():
             self._disconnected.set_result(None)
 
+    def _outgoing_call(self, func: Callable) -> Callable[..., Awaitable]:
```
frederikaalund, line 230: Well done with this decorator. :+1: 

One nitpick: The outgoing call may return something different from `None` (e.g., `subscribe` returns `int`). Let's preserve the return type via a `TypeVar`. Similar to what [they do in this example from the standard library](https://docs.python.org/3/library/typing.html#typing.ParamSpec). Note that we don't have `ParamSpec` yet (It's in 3.10) but we can still preserve the return type.

asyncio_mqtt/client.py
frederikaalund: Also, let's add [`functools.wraps` to preserve the attributes of the original function](https://docs.python.org/3/library/functools.html#functools.wraps) (`__name__`, `__doc__`, etc.).

asyncio_mqtt/client.py
```diff
@@ -218,6 +227,18 @@ async def force_disconnect(self) -> None:
...
+        async def decorated(*args: Any, timeout: int = 10, **kwargs: Any) -> None:
+            if not self._outgoing_calls_sem:
+                func(*args, timeout=timeout, **kwargs)
```
frederikaalund, line 233: Just `return await func(...)` to preserve the returned value. :+1:  Also note the missing `await`.

asyncio_mqtt/client.py
```diff
@@ -218,6 +227,18 @@ async def force_disconnect(self) -> None:
...
+                return
+
+            async with self._outgoing_calls_sem:
+                func(*args, timeout=timeout, **kwargs)
```
frederikaalund, line 237: Use `return await func(...)` here as well

frederikaalund (changes requested):
Thanks for this PR. :+1: Please have a look at my comments in the review. Let me know if anything is unclear, or, if you have other suggestions. :)

frederikaalund:
I'd also like a comment somewhere along these lines:

```python
# TODO: Simplify the logic that surrounds `self._outgoing_calls_sem` with `nullcontext` when we
# support Python 3.10 (`nullcontext` becomes async-aware in 3.10).
# See: https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext
```

asyncio_mqtt/client.py
bachya: I included it to more explicitly mirror the method declarations that it's wrapping – example:

https://github.com/sbtinstruments/asyncio-mqtt/blob/2a352ba87c35d4c4ee22fa4420823828e1fff025/asyncio_mqtt/client.py#L243

That said, happy to simplify if that's what you'd prefer.

asyncio_mqtt/client.py
bachya: What about https://github.com/sbtinstruments/asyncio-mqtt/pull/101/commits/7274933734c3b1568e662e3b1934e8171295d86e (with an explicit explanation of what types are possible)?

frederikaalund: Good question. :) I prefer `TypeVar` over `Union` due to type scope. See https://stackoverflow.com/a/58903907/554283 for reference. Let me know if it makes sense or not.

asyncio_mqtt/client.py
frederikaalund: Let's simplify it away then. 👍 Shorter upgrade path towards `ParamSpec` in 3.10. Similar "type scope" arguments as for the return type (see the other comment).

asyncio_mqtt/client.py
bachya: https://github.com/sbtinstruments/asyncio-mqtt/pull/101/commits/b6e5b680f88f38742c057fe9cb3f13e726ffceb2?

frederikaalund: Something like this (please test it first):

```python
T = TypeVar("T")

def _outgoing_call(self, func: Callable[..., Awaitable[T]]) -> Callable[..., Awaitable[T]]:
    @wraps(func)
    async def _decorated(*args: Any, **kwargs: Any) -> T:
        ....
    return _decorated
```

bachya: This shows how badly I understand TypeVars... 😆 `mypy` doesn't complain about this, so I'm assuming it's good.

frederikaalund: Don't worry about it. It is always a pain to add typing to a decorator function. :))

asyncio_mqtt/client.py
```diff
@@ -61,6 +64,8 @@ class ProtocolVersion(IntEnum):
 # TODO: This should be a (frozen) dataclass (from Python 3.7)
 # when we drop Python 3.6 support
 class Will:
+
+    OutgoingType = TypeVar("OutgoingType", None, int)
```
frederikaalund, line 68: I think this is an unintended left-over. :)

frederikaalund:
I added one last comment. When that is done this PR looks good to me. 👍 Thanks for the patience with my pedantic obsession with typing. 😅

frederikaalund:
Looks good to me now. Thank you for your contribution to asyncio-mqtt. 👍

bachya:
Thank you so much for your support! So I can plan: when are you thinking about cutting a new release?

frederikaalund:
First thing tomorrow. I'll keep the notification e-mail about this PR in my inbox to remind me.

frederikaalund:
I just uploaded v0.12.0 to PyPI.

PR #107, opened 2022-03-09
Adds LICENSE to built tarball

stewarthaines:
* adds setup.cfg with license_files

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[metadata]
-license_files = LICENSE
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[metadata]
+license_files = LICENSE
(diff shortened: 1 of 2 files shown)


frederikaalund:
Hi Stewart. Thanks for opening this pull request. Let me have a look. :)

I like the simplicity of this solution. :+1: 
I think that it's possible to simplify it even more and add the following to `setup.py`:

    license_files=["LICENSE"]

This way, we avoid another file in the root of the repo. See https://stackoverflow.com/a/66443941/554283 for reference.

stewarthaines:
Thanks Frederick.

That seems to do the job as well.

frederikaalund:
Looks good to me. :) Thank you for your contribution to asyncio-mqtt. :+1:

PR #114, opened 2022-05-26
client: Fix thread safety in socket close callback

AdmiralNemo:
The `_on_socket_close` method can be called from a different thread than the
running event loop if an error occurs while the connection is being
established.  This can happen, for example, if the TCP connection is
established, but the TLS or WebSocket handshake fails.  It is not safe to
cancel the task from another thread.  Instead, the cancellation should be
scheduled to run as soon as the event loop is idle.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -534,8 +534,7 @@ class Client:
         if fileno > -1:
             self._loop.remove_reader(fileno)
         if self._misc_task is not None and not self._misc_task.done():
-            with suppress(asyncio.CancelledError):
-                self._misc_task.cancel()
+            self._loop.call_soon_threadsafe(self._misc_task.cancel)
 
     def _on_socket_register_write(
         self, client: mqtt.Client, userdata: Any, sock: socket.socket


frederikaalund:
Hi Dustin. Thanks for this pull request. 👍  Let me have a look.

First of all, good catch! This is something that we need to fix. I only have one comment:

```python
with suppress(asyncio.CancelledError)  # This `with` statement has no effect
    self._loop.call_soon_threadsafe(self._misc_task.cancel)  # This returns immediately
```

The `self._loop.call_soon_threadsafe` call schedules the given callback and exits immediately. Therefore, the `suppress(asyncio.CancelledError)` context manager has no effect. This leaves us with the question: What happens if the `self._misc_task.cancel` call raises an exception when it runs? I actually don't know the answer to that. There are two options:

 1. asyncio discards the exception. Then, we don't need to do anything (we can simply remove the `suppress(asyncio.CancelledError)` part).
 2. asyncio raises the exception somewhere (e.g., at the `asyncio.run`). Then, we need to wrap `self._misc_task.cancel` with the `suppress(asyncio.CancelledError)` context manager and schedule that wrapper function instead.

We need to resolve this question before we merge this PR. Are you up for that? :)

AdmiralNemo:
I don't think calling [`Task.cancel`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) can raise `CancelledError`, so it does seem like the `suppress` context manager is not needed.

frederikaalund:
I agree (had a look at [CPython the source](https://github.com/python/cpython/blob/main/Lib/asyncio/tasks.py#L188)).

Thank you for your contribution to asyncio-mqtt. 👍

PR #121, opened 2022-08-03
Bugfix: Update client.py

miili:
Catching right exception

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -226,7 +226,7 @@ class Client:
             # get_running_loop is preferred, but only available in python>=3.7
             try:
                 loop = asyncio.get_running_loop()
-            except AttributeError:
+            except RuntimeError:
                 loop = asyncio.get_event_loop()
 
             # [3] Run connect() within an executor thread, since it blocks on socket


frederikaalund:
Hi miili. Thanks for opening this pull request. Let me have a look. :)

The intention of the `AttributeError` is to have a fallback for python 3.6 and below. Specifically, because Python 3.7 added `get_running_loop`.

In python 3.6, the following occurs:
```python
>>> asyncio.get_running__loop()
  File "<stdin>", line 1, in <module>
AttributeError: module 'asyncio' has no attribute 'get_running_loop'.
```
This is exactly what we want. In the exception handler for this `AttributeError`, we fall back to the older `asyncio.get_event_loop`.

---

Are you sure that you have a running event loop when you call connect? Because if you don't, then you get a `RuntimeError`. E.g.:

```python
>>> asyncio.get_running_loop()
  File "<stdin>", line 1, in <module>
RuntimeError: no running event loop
```

empicano:
I think we can close this. In `0.14.0` we dropped Python 3.6 support, and the `try/except` of this PR was removed. If there's anything left unclear, simply open a new issue 😋

Issue #122, opened 2022-08-09
asyncio.exceptions.InvalidStateError when cancellation is between command trigger and receive of callback


aha79:
In my application using asyncio-mqtt I received an InvalidStateError in this [`set_result` call](https://github.com/sbtinstruments/asyncio-mqtt/blob/6b02071227635fa532698b55c5159755f4e411b2/asyncio_mqtt/client.py#L492)

This happend after my application terminated itself during startup (due to invalid config file, a reason not releated to asyncio-mqtt). 
So the InvalidStateError  was raised in course of processing of another exception (not related to mqtt).

This exception handling lead to immediate cancellation of all tasks of the application. However, the mqtt module was still in its startup with various subscribe events inplace and the _on_subscibe callback not yet called. This caused the Futures in 'self._pending_subscribes' being cancelled, while the callback was not received. When the callback was finally executed the future was already cancelled and the 'set_result' call raised InvalidStateError.

As this is a hard to reproduce issue, I have no test program. However, I guess, what happens is clear.

I would suggest to guard all 'fut.set_result' and 'fut.set_exception' calls with an if:

    if not fut.cancelled(): 
         fut.set_result(...)  # or fut.set_exception()

This will avoid the InvalidStateError in case the future was cancelled right during waiting for a callback.

Even though I only observed it in a on_subscribe callback, I guess it would make sense to introduce this guard to all callbacks.
 
Do you think that makes sense?

frederikaalund:
Hi aha9, thanks for opening this issue. :+1: Let me have a look.

Hmm, I feared that I had missed a couple of `fut.cancelled` guards somewhere. Until now, I add these guards in response to issues. At this point, however, I concur that it makes sense to simply guard all `fut.set_result` calls. It's impratical to figure out when the guards are actually necessary or not so it's better to just guard everything. Better safe than sorry.

> Do you think that makes sense?

In short: Yes! Do you want to make a PR? :)

aha79:
Yes sure. Let me give it a try.

aha79:
Upon closer look into the code I discovered the cancelled guards you mentioned. Actually you had them everywhere except for the `set_result` call in the _on_subscribe callback :)

I added a guard there and also for two 'set_exception' calls. They face the same problem and it probably will not harm to have a guard there as well

frederikaalund:
Good point about the `set_exception` calls. 👍 Thanks for the PR. I merged it just now. :)

PR #123, opened 2022-08-09
Fix race conditions that may cause `InvalidStateError`.

aha79:
This fixes https://github.com/sbtinstruments/asyncio-mqtt/issues/122


frederikaalund:
LGTM. 👍 Thank you for your contribution to asyncio-mqtt! 🎉

Issue #14, opened 2020-08-10
Add the possibility to use tls_set() in addition to tls_context


svinz:
To ease the use of certificates for authentication, it would be really nice if the tls_set function was implemented in asyncio-mqtt client.py. 

This would be a nice feature for newbies (like me)

frederikaalund:
Hi svinz, thanks for raising this issue. Let me have a look. :)

Indeed, the `paho.mqtt.Client.tls_set` functionality is not currently part of asyncio-mqtt. I'm definitely open for adding it. Preverably, as part of the `asyncio_mqtt.Client.__init__`'s keyword arguments. Do you want to add it via a pull request?

We do already have the `tls_context` keyword argument that you can use to set up TLS. Maybe that can work for you as well? :)

Let me know your thoughts.

svinz:
Yeah, I saw the `tls_context` in `client.py`. 
After having a look at [tls_set in paho.mqtt.client](https://github.com/eclipse/paho.mqtt.python/blob/d7592a64b55469035ab66160919561c30ae5f645/src/paho/mqtt/client.py#L760), it seems rather ok to implement this.

I made a small test by copy&paste from `paho.mqtt.python.client.tls_set` into `asyncio_mqtt.Client.__init__` and it seems to work. I can try to make a PR tomorrow, though I'm not very proficient at this.

Alex

frederikaalund:
Glad to hear that. :)

As far as I can see, Paho's `Client.tls_set` is just a convenience method that creates an `ssl.SSLContext` and then calls `Client.tls_set_context`. Basically a convenience method. Seems fair enough to add this convenience to asyncio-mqtt as well. :)

Let me offer some design suggestions: Add this functionality via keyword arguments to the `Client` constructor. Maybe something like a `tls_tuple` keyword argument. E.g.:

```python
from asyncio_mqtt import TlsTuple, Client

tls_tuple = TlsTuple(ca_certs=None, certfile=None, keyfile=None, cert_reqs=None, tls_version=None, ciphers=None)
client = Client("localhost", tls_tuple=tls_tuple)
```

Where `TlsTuple` is a new class that basically just structures the arguments together. You can use Python's `collections.namedtuple` to create it (ideally, we would use `dataclasses.dataclass`, but that is 3.7 only, and we want to support 3.6 for now).

In turn, the we can them simply forward the given tuple as follows (inside the `__init__` function)

```python
if tls_tuple is not None:
[...]

PR #126, opened 2022-08-30
use tls_set() in addition to tls_context

Sohaib90:
I use the comments about the design of adding `tls_set` to `asyncio-mqtt` from https://github.com/sbtinstruments/asyncio-mqtt/issues/14#issuecomment-671780413 and try to implement in the light of that. 

My implementation allows the use of a `TLSParameters` object to invoke `tls_set()` function of paho-mqtt client in asyncio-mqtt. Let me know if it makes sense. I would also like some pointers in regards to testing this

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -87,6 +87,7 @@ class Will:
 class TLSParameters:
     def __init__(
         self,
+        *,
         ca_certs: Optional[str] = None,
         certfile: Optional[str] = None,
         keyfile: Optional[str] = None,
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -139,7 +139,7 @@ class Client:
         logger: Optional[logging.Logger] = None,
         client_id: Optional[str] = None,
         tls_context: Optional[ssl.SSLContext] = None,
-        tls_set_params: Optional[TLSParameters] = None,
+        tls_params: Optional[TLSParameters] = None,
         protocol: Optional[ProtocolVersion] = None,
         will: Optional[Will] = None,
         clean_session: Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


asyncio_mqtt/client.py
```diff
@@ -83,6 +83,27 @@ def __init__(
         self.retain = retain
         self.properties = properties
 
+# TLS set parameter class
+class TLSParameters:
+    def __init__(
+        self,
```
frederikaalund, line 89: Change to `__init__(self, *, ca_certs, ...)`. This way, we force the user to supply the keyword argument, which is a good thing IMHO when there are this many options. 👍

frederikaalund (changes requested):
Thanks for opening this PR. 👍 I only have one minor comment (see the review). Other than that, it LGTM. Well done!

asyncio_mqtt/client.py
Sohaib90: Since there are default values to these parameters (even in paho-mqtt), then is it okay to add `*` which will force the user to give the values even if it is none? I looked up the usage of `*` in function calls, which forces the user to give value to an argument that has no default value.

asyncio_mqtt/client.py
```diff
@@ -83,6 +83,27 @@ def __init__(
...
+        ca_certs: Optional[str] = None,
+        certfile: Optional[str] = None,
+        keyfile: Optional[str] = None,
+        cert_reqs: Optional[ssl.VerifyMode] = ssl.CERT_REQUIRED,
```
Sohaib90, line 93: Also, one more change I just saw in paho-mqtt is that I dont have to provide the default values, these should be `None` as well. So I will change this as well

asyncio_mqtt/client.py
frederikaalund: Yes, the `*` simply forces the user to use keyword arguments instead of positional arguments. All other semantics stay the same (including defaults). 👍

asyncio_mqtt/client.py
frederikaalund: Well spotted. 👍 I missed that one.

frederikaalund:
All that's left is the `*` to force the user to supply keyword arguments. 👍

asyncio_mqtt/client.py
Sohaib90: Oh makes sense. Added and committed. Please review it :)

asyncio_mqtt/client.py
```diff
@@ -117,6 +139,7 @@ def __init__(
         logger: Optional[logging.Logger] = None,
         client_id: Optional[str] = None,
         tls_context: Optional[ssl.SSLContext] = None,
+        tls_set_params: Optional[TLSParameters] = None,
```
frederikaalund, line 142: One last nitpick here, and then I promise it's good to merge. 😅 Can we change the name of the parameter from `tls_set_params` to simply `tls_params`? Thanks.

Sohaib90: Of course. That makes more sense now that I think of it

frederikaalund:
Thank you for your patience and thank you for your contribution to asyncio-mqtt. 👍

Sohaib90:
Learned a lot and thank you for the library. It is a great help to so many :)

PR #127, opened 2022-09-01
paho.mqtt.proxy_set() - support

Sohaib90:
This commit adds the functionality of paho-mqtt's `proxy_set()` to asyncio-mqtt

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -104,7 +104,25 @@ class TLSParameters:
         self.ciphers = ciphers
         self.keyfile_password = keyfile_password 
         
-
+# Proxy parameters class 
+class ProxySettings:
+    def __init__(
+        self,
+        *, 
+        proxy_type: int = None,
+        proxy_addr: str = None,
+        proxy_rdns: Optional[bool] = True,
+        proxy_username: Optional[str] = None,
+        proxy_password: Optional[str] = None
+    ):
+        self.proxy_args = {
+            "proxy_type" : proxy_type,
+            "proxy_addr" : proxy_addr,
+            "proxy_rdns" : proxy_rdns,
+            "proxy_username" : proxy_username,
+            "proxy_password" : proxy_password 
+        }
+        
 
 # See the overloads of `socket.setsockopt` for details.
 SocketOption = Union[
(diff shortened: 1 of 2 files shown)


asyncio_mqtt/client.py
```diff
@@ -104,7 +104,25 @@ def __init__(
         self.ciphers = ciphers
         self.keyfile_password = keyfile_password 
         
-
+# Proxy parameters class 
+class ProxySettings:
+    def __init__(
+        self,
+        *, 
+        proxy_type: int = None,
```
frederikaalund, line 112: You forgot the `Optional` part here. :)

frederikaalund (changes requested):
Thanks for opening this PR. :+1: Looks good to me. I only have one minor comment about typing (see the review).

asyncio_mqtt/client.py
Sohaib90: If you look at the description here: https://github.com/eclipse/paho.mqtt.python/blob/225ab3757f6818ba85eb80564948d1c787190cba/src/paho/mqtt/client.py#L875

It says 
```
(Required)
proxy_type: One of {socks.HTTP, socks.SOCKS4, or socks.SOCKS5}
proxy_addr: IP address or DNS name of proxy server
(Optional)
proxy_rdns: boolean indicating whether proxy lookup should be performed
    remotely (True, default) or locally (False)
proxy_username: username for SOCKS5 proxy, or userid for SOCKS4 proxy
proxy_password: password for SOCKS5 proxy
```

In line with that shouldnt it be like this, or should I change it to Optional?

frederikaalund: Ah, good point. :+1:  If those arguments are _required_ then let's not use default valures (`= None`) for them.
Looking at the [paho implementation](https://github.com/eclipse/paho.mqtt.python/blob/225ab3757f6818ba85eb80564948d1c787190cba/src/paho/mqtt/client.py#L3477), we could have a signature like this:

```python
def __init__(
    self,
    *, 
    proxy_type: int,
    proxy_addr: str,
    proxy_rdns: Optional[bool] = True,
    proxy_username: Optional[str] = None,
    proxy_password: Optional[str] = None
): ...
```

This way, the user must supply the `proxy_type` and `proxy_addr` arguments.

> Also do you think it is a good idea to add some documentation so that the users know that the functionality exists (e.g a small example) in the README.md in another PR?

Yes, that's a good idea. :+1: 
Specifically, we should let people know that this whole proxy is an ["extra" feature (even in paho) that requires the `PySocks` dependency](https://github.com/eclipse/paho.mqtt.python/blob/master/setup.py#L18).

Sohaib90: I have fixed it and committed it as well. 
Alright. I will work on the documentation and create a PR for that

frederikaalund:
LGTM and merged. :+1: Thank you for your contribution to asyncio-mqtt (again). :)

Sohaib90:
Thank you for the opportunity (again) :D

PR #129, opened 2022-10-05
Fix 'asyncio_mqtt' has no attribute 'TLSParameters' error

empicano:
Hi Frederik 😊

Using the current code straight from GitHub leads to a `'asyncio_mqtt' has no attribute 'TLSParameters'` error on setting TLS configuration. I think this is because the import in `__init__.py` is missing. #126 is the relevant commit.

In case you're looking for help, I'll try to set some time apart to help out, as I'm actively using `asyncio-mqtt` at the moment.

### DIFF
--- a/asyncio_mqtt/__init__.py
+++ b/asyncio_mqtt/__init__.py
@@ -1,11 +1,14 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from .error import MqttError, MqttCodeError
-from .client import Client, Will, ProtocolVersion
+from .client import Client, Will, ProtocolVersion, TLSParameters
 from .version import __version__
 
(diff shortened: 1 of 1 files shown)


frederikaalund:
Hi Felix, thanks for opening this PR. Let me have a look. :)

Ah, nice find! This is an oversight on my part. Thanks for the fix. :+1: 

> In case you're looking for help, I'll try to set some time apart to help out, as I'm actively using asyncio-mqtt at the moment.

Greatly appreciated. :) Feel free to help out wherever you deem necessary. I'll try to review issues/PRs/discussions in a timely fashion. If I don't respond within a day, just ping me. :+1:

PR #128, opened 2022-09-01
Added documentation for tls_set in README.md

Sohaib90:
I just wanted to ask for improvements and suggestions when adding the recent changes to the README.md
If this is alright (for now, just tls_set example added) I can add the proxy settings as well

PR #130, opened 2022-10-05
Fix missing ProtocolVersion import in TLS configuration example

empicano:
The TLS configuration example in the README fails to run as is because the `ProtocolVersion` is not imported. I added that small change (line 153), relevant PR is #128. I also formatted the `README` with `Prettier` for some more consistency, I hope that's fine with you, otherwise, I'll revert it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@
 
 Write code like this:
 
-
 ##### Subscriber
+
 ```python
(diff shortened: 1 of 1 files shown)


frederikaalund:
Thanks again. :)

I use Prettier for other projects, so I'm perfectly fine with it. :+1: 

~Frederik

Issue #105, opened 2022-02-24
Is it necessary to create a topic filter?


roipoussiere:
Here is the code sample from the readme:

```py
async with Client("test.mosquitto.org") as client:
    async with client.filtered_messages("floors/+/humidity") as messages:
        await client.subscribe("floors/#")
        async for message in messages:
            print(message.payload.decode())
```

I don't understand why both `client.filtered_messages("floors/+/humidity")` and `client.subscribe("floors/#")` are required here, since mqtt topics are already filtered by the mqtt broker. As far as I understand, we could just subscribe to `client.subscribe("floors/+/humidity")`.

So I would like to understand if the `client.filtered_messages` is really necessary and why/when to use it.

frederikaalund:
Good question! :)

Indeed, you can just use `subscribe("floors/+/humidity")` combined with `unfiltered_messages`. This way, you get all messages that match the `floors/+/humidity` topic.

For some use cases, you want to subscribe to a general topic such as `floors/#` but handle the messages differently based on the actual topic of each received message. For this purpose, you use `filtered_messages("floors/basement/#")`, `filtered_messages("floors/2nd-floor/#")`, and then a single `unfiltered_messages` for the rest (everything from `floors/#` that isn't explicitly `floors/basement` or `floors/2nd-floor`). Does it make sense?

roipoussiere:
> For some use cases, you want to subscribe to a general topic such as `floors/#` but handle the messages differently based on the actual topic of each received message.

Hmm, I don't really understand because we could subscribe to both `floors/basement/#` and `floors/2nd-floor/#` independently and to tasks according to our needs. But I'm new to mqtt and I might miss some use cases.

I don't need filters anyway, but if I remove the line with the `client.filtered_messages`, I don't know how to handle incoming messages... Could you provide a few lines of code with a topic subscription without a filter?

frederikaalund:
> Hmm, I don't really understand because we could subscribe to both floors/basement/# and floors/2nd-floor/# independently and to tasks according to our needs

True, and you get all of said messages interleaved. We use `filtered_messages` to deinterleave the messages. Of course, if your use case doesn't require deinterleaving, then just go ahead and use `unfiltered_messages`. :+1: 

> I don't need filters anyway [...]

Great, just use `unfiltered_messages`. :) Something like:

```python
async with Client("test.mosquitto.org") as client:
    async with client.unfiltered_messages() as messages:  # Receive me all messages
[...]

Issue #117, opened 2022-06-23
Provide same utility and simplicity without context manager


robin-blanchard:
Hello!

I am building a library around many asyncio connectors. And the structure I've chosen made me prefer to use connectors that I can connect and disconnect at different places in my code. This choice prevents me from easily using `asyncio-mqtt.Client` context manager.  
While the `__aenter__` method can easily be replaced by `.connect()`, the `__aexit__` provides much more functionality than the `.disconnect()` method. 
Would it be okay to wrap the content of the `__aexit__` method inside a new method ? Possibly naming it `disconnect` and renaming the current `disconnect` by `gracefully_disconnect` ?

frederikaalund:
Hi Robin, thanks for opening this issue. :)

I understand your trouble with this whole context manager thing. It is a paradigm shift compared to a `connect`/`disconnect` function pair. It can be difficult to mix the two paradigms. That's what you're experiencing here.

Long story short, it was an API design mistake on my part to expose `connect`/`disconnect`/`force_disconnect` in the first place. With that in mind, I personally vote against a new `gracefully_disconnect` function. There is already enough confusion with the existing `disconnect`/`force_disconnect`. I'd rather remove all these functions and rely solely on the context manager semantics. Unfortunately, it's a little too late now for a breaking API change. That's on me.

---

As a workaround, you can call `__aenter__` and `__aexit__` directly. E.g.:

```python
await client.__aenter__()  # connect
...
await client.__aexit__(None, None, None)  # workaround
```

Moreover, you can create a small wrapper around `asycio_mqtt.Client` that does just that:

```python
class RobinsClient(asyncio_mqtt.Client):
    async def smart_disconnect(self) -> None:
        # HACK: We use the existing disconnect logic in asyncio-mqtt with
        # a direct call to `__aexit__`. This may break if asyncio-mqtt changes the implementation
        # of `__aexit__`. Make sure to check that whenever you upgrade the version of asyncio-mqtt.
        await self.__aexit__(None, None, None)
```

robin-blanchard:
Hello Frederik,

Thanks for your quick reply.
I understand and agree on the fact that there is already confusion with `disconnect`/`force_disconnect`.

I indeed was planning on using the `__aexit__` method directly, thank you for your help !

couling:
[...]

PR #131, opened 2022-10-07
Simplify and extend usage documentation

empicano:
Hi Frederik!

The _Advanced usage_ section may be a bit overwhelming at first sight. I was thinking about dissecting it into smaller sections/examples that each explains only a single concept e.g. _Reconnecting_ (see lines 41-56).

This makes it more in line with the TLS and Proxying sections and would make it easier to add/remove/edit the examples in the future. I'd also like to add a section about topic filters, closing #105.

Let me know what you think 😊

-- Felix

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The whole thing is less than [700 lines of code](asyncio-mqtt/client.py).
   - [Configuring the client](#configuring-the-client)
   - [Reconnecting](#reconnecting)
   - [Sharing the connection](#sharing-the-connection)
-  - [Side by side with async web frameworks](#side-by-side-with-async-web-frameworks)
+  - [Listening without blocking](#listening-without-blocking)
   - [Topic filters](#topic-filters)
   - [TLS](#tls)
   - [Proxying](#proxying)
@@ -124,7 +124,7 @@ asyncio.run(main())
 
 ### Topic filters
 
-Let's take the example from the beginning again, but this time with messages in both `measurements/humidity` and `measurements/temperature`. You want to receive both types of measurements, but handle them differently. asyncio-mqtt has topic filters to make this easy:
+Let's take the example from the beginning again, but this time with messages in both `measurements/humidity` and `measurements/temperature`. You want to receive both types of measurements but handle them differently. asyncio-mqtt has topic filters to make this easy:
 
 ```python
 import asyncio
@@ -194,7 +194,10 @@ asyncio.run(main())
 
 ### Sharing the connection
 
-In many cases you'll want to send and receive messages in different locations in your code. You could create a new client each time, but firstly this not very performant, and secondly you'll use a lot more network bandwith.
+In many cases, you'll want to send and receive messages in different locations in your code. You could create a new client each time, but
+
+1. this is not very performant, and
+2. you'll use a lot more network bandwidth.
 
 You can share the connection by passing the `Client` instance to all functions that need it:
 
@@ -222,48 +225,68 @@ asyncio.run(main())
 
 **Caveats:**
 
-Most web frameworks take control over the "main" function, which makes it difficult to figure out where to create and connect to the `Client`. With e.g. FastAPI you can share a connection via its dependency injection.
+Most web frameworks take control over the "main" function, which makes it difficult to figure out where to create and connect to the `Client`.
 
-### Side by side with async web frameworks
-
-If you run the basic example for subscribing and listening for messages, you'll notice that the program doesn't finish until you stop it. If you want to use asyncio-mqtt side by side with a web framework you don't want the whole program to block after starting your listener. You can use asyncio's `create_task` for this. This is similar to starting a new thread without `join`ing it in a multithreaded application.
+Some frameworks like [Starlette](https://github.com/encode/starlette) directly support lifespan context managers, with which you can safely set up a global client instance:
 
 ```python
 import asyncio
 import asyncio_mqtt as aiomqtt
 import starlette.applications
-import starlette.responses
-import starlette.routing
 
 
-async def startup():
-    async def listen():
-        async with aiomqtt.Client("test.mosquitto.org") as client:
-            async with client.unfiltered_messages() as messages:
-                await client.subscribe("measurements/#")
-                async for message in messages:
-                    print(message.payload)
-
-    # wait for messages in (unawaited) asyncio task
-    loop = asyncio.get_event_loop()
-    loop.create_task(listen())
-
-    # this will still run!
-    print("Magic!")
+client = None
 
 
-async def index(request):
-    # and thus requests can be handled
-    return starlette.responses.JSONResponse({"hello": "world"})
+@contextlib.asynccontextmanager
+async def lifespan():
+    global client
+    async with aiomqtt.Client("test.mosquitto.org") as c:
+        client = c
+        yield
 
 
 app = starlette.applications.Starlette(
-    routes=[starlette.routing.Route(path="/", endpoint=index, methods=["GET"])],
-    on_startup=[startup],
+    routes=[],
+    lifespan=lifespan,
 )
 ```
 
-Starlette is only used as an example here. This works similarly with other frameworks.
+FastAPI, which is built upon Starlette, doesn't expose that API yet, but there's [an open PR](tiangolo/fastapi#2944) to add it. In the meantime, you can share a connection via FastAPI's dependency injection.
+
+### Listening without blocking
+
+If you run the basic example for subscribing and listening for messages, you'll notice that the program doesn't finish until you stop it. If you want to run other code after starting your listener (e.g. handling HTTP requests) you don't want the execution to block.
+
+You can use asyncio's `create_task` for this. The concept is similar to starting a new thread without `join`ing it in a multithreaded application.
+
+```python
+import asyncio
+import asyncio_mqtt as aiomqtt
+
+
+async def listen():
+    async with aiomqtt.Client("test.mosquitto.org") as client:
+        async with client.unfiltered_messages() as messages:
+            await client.subscribe("measurements/#")
+            async for message in messages:
+                print(message.payload)
+
+
+async def main():
+    # wait for messages in (unawaited) asyncio task
+    loop = asyncio.get_event_loop()
+    task = loop.create_task(listen())
+    # this will still run!
+    print("Magic!")
+    # if you don't await the task here the program will simply finish.
+    # however, if you're using an async web framework you usually don't have to await
+    # the task, as the framework runs in an endless loop.
+    await task
+
+
+asyncio.run(main())
+```
 
 ### TLS
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,24 @@ asyncio-mqtt combines the stability of the time-proven [paho-mqtt](https://githu
 - Fully type-hinted
 - Did we mention no more callbacks?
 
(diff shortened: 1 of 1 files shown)


frederikaalund:
Hi Felix. :)

Thanks again for taking your time to improve asyncio-mqtt. It's really appreciated. 👍 

Great idea to make the readme more approachable/comprehensible. In general, we lack on the documentation/examples side. I welcome all improvements in that area.

As for your specific suggestion, I think it makes great sense to split the readme up into smaller sections. Makes it easier to digest a given topic. Maybe this grows into an actual documentation (e.g., "Read the Docs" style) in the future.

So by all means, go ahead. :)

~Frederik

frederikaalund:
Your contributions so far are on point and in line with my own thinking/direction. If you're up for it, you're very welcome to join the asyncio-mqtt team. :) There are no strings attached, you just contribute/maintain asyncio-mqtt whenever you feel like it (or not at all if you grow tired of it).

You'll get direct read/write rights to the repo, so you can change the code directly without PRs. Of course, you're welcome to keep the PR workflow and I'll gladly review/comment on the PRs. For minor/non-breaking changes, it's just easier to edit the repo directly.

~Frederik

empicano:
> As for your specific suggestion, I think it makes great sense to split the readme up into smaller sections. Makes it easier to digest a given topic.

Great! I'll write something down in the next few days :)

> Your contributions so far are on point and in line with my own thinking/direction. If you're up for it, you're very welcome to join the asyncio-mqtt team. :) There are no strings attached, you just contribute/maintain asyncio-mqtt whenever you feel like it (or not at all if you grow tired of it).

Sure, I'm up for it, thanks for the trust! I'll keep with the PR workflow for most things I think, I like when someone looks over it, then I can learn from the comments :)

frederikaalund:
> Great! I'll write something down in the next few days :)

Sounds good! I'll be ready to review the PR.

> Sure, I'm up for it, thanks for the trust! I'll keep with the PR workflow for most things I think, I like when someone looks over it, then I can learn from the comments :)

Exellent! I'll add you right away. Welcome aboard. :)

frederikaalund:
Looking good so far. 👍 I especially like the new `### Configuring the client` section. It neatly showcases the available functionality, Likewise, it's nice with several, specialized examples instead of a single, "advanced" example.

Could merge as is but let me know when you want to do so. :)

empicano:
Great that you like it! 🎉 

---

I stumbled on two things while writing this down. **(1.)** In #105 you said that "we use `filtered_messages` to deinterleave [...] messages". I still don't really understand the benefit of the example in the README over:

```python
    async with client.unfiltered_messages() as messages:
        await client.subscribe("measurements/#")
            if message.topic == "measurements/humidity":
                # do something humidity specific
            if message.topic == "measurements/temperature":
                # do something temperature specific
```

Is it that with `filtered_messages` messages of different topics can be processed concurrently? Or that the different filters can be changed afterward more easily? Or that the filters can include special characters like `measurements/+/something` (though that could be easily implemented differently)?

The current way seems overly complicated to me at the moment, so I feel like I'm not seeing something 😄 

---

**(2.)** is that I don't understand how to share an MQTT connection between functions. In #82 (which was sadly handled very childishly by the issue author) you said that sharing the client between different files is an anti-pattern.

I searched around quite a bit, but I didn't find anything for "sharing context managers", "sharing database connection via context managers", or similar. Maybe I'm searching in the wrong direction. Could you give me some pointers on how I can find more information on this? Or is the idea to open a new connection for each message?

I'd like to add a section about this to the README as well.

frederikaalund:
Great questions. Let me try to answer them. :)

Indeed, it's a bit strange that we have both `filtered_messages` vs `unfiltered_messages`. It's because paho-mqtt has both. Personally, I would just have `unfiltered_messages` (and simply call it `messages`) and then do the filtering manually (like you do with `if`s in your example code).
asyncio-mqtt is *just* a wrapper around paho-mqtt. In that spirit, I decided to not change the interface too much (and keep the `filtered` functionality). Maybe that wasn't a wise decision. 😅 
As far as I know, paho-mqtt basically reimplements the topic wildcards (`+` and `*`) on the client for `filtered_messages`. So it does provide some convenience over simple `if`s. Enough convenience to warrant the existence of `filtered`? Personally, I don't really think  so but the paho-mqtt developers apparently thought so. :)

---

Yeah, #82 did have some merit because we don't have examples to show how to do it proper. The conversation just got a bit off track.
To share a client between multiple functions, just pass the `Client` instance to the function! That's all there is to it. I usually have the `async with Client()` in the "main" function and then pass said `Client` to all other functions that want to subscribe/publish.

However, there are two problems with this approach:
 1. Some frameworks take control over the "main" function, which makes it difficult to figure out where to create/enter the `Client`.
 2. What if the connection breaks? There is no reconnect functionality.

Frameworks such as FastAPI solve (1) via their dependency system. We can give an example for FastAPI just to show how.
Problem (2) is more difficult. It requires more fundamental change. Either we make `Client` itself "auto-reconnect" internally, or, we create a wrapper around `Client` (say, `ReconnectingClient`) that does so. I neither case, it requires some dev work (PRs are welcome :D)

---

I hope that answers your questions. Those were good questions that showcases some of the rough edges in asyncio-mqtt. Keep them coming. 👍

empicano:
**(1.)** I get that, staying close to paho.mqtt makes the asyncio-mqtt much leaner. Personally, I don't think that's worth having to deal with `AsyncExitStack` for multiple filters. paho.mqtt does indeed reimplement the topic wildcards in [`paho/mqtt/matcher.py`](https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/matcher.py). The logic is pretty simple. If you're ok with it, I'll take a stab at this in a while, for something like:

```python
        await client.subscribe("measurements/#")
            if matches(payload.topic, "measurements/+/something"):
                print(message.payload)
```

---

**(2.)** That makes sense! I'm actually using asyncio-mqtt inside a web framework, that's why I am struggling. Using FastAPI's dependency injection to share context is pretty neat, for other frameworks it sadly doesn't work that easily.

I read in e.g. #117 that you're leaning against using `connect/disconnect` directly. I think it could be a nice solution for other web frameworks, usually, there's some sort of `startup/shutdown` procedure. It doesn't seem too unusual to have both either, see [`encode/databases`](https://www.encode.io/databases/connections_and_transactions/).

Automating reconnection would be pretty cool, I already considered that as one of the next things to do 👍

empicano:
I now also added an example of how to start a listener inside a web framework without blocking the rest of the code. I'm happy about all feedback! 😊

frederikaalund:
> **(1.)** [...] The logic is pretty simple. If you're ok with it, I'll take a stab at this in a while, for something like:
> 
> ```python
> async with Client("test.mosquitto.org") as client:
>     async with client.messages() as messages:
>         await client.subscribe("measurements/#")
>         async for message in messages:
>             if matches(payload.topic, "measurements/+/something"):
>                 print(message.payload)
> ```

That's much cleaner and more approachable than the current API. Well done. 👍 Feels more natural to split up the semantics (iterating messages and filtering messages) into separate operations. I also see how we can add this in a backwards-compatible way. I welcome a PR with that change. :)

You could also choose to do a `Message.matches` like so:

     if message.matches("measurements/+/something"):

Of course, that'll mean that we have to create a `asyncio_mqtt.MqttMessage` class that wraps `mqtt.MQTTMessage`. I just put the idea here as an alternative. You decide which one that you want to proceed with (`matches`, `Message.matches`, or something new). 👍 

> **(2.)** That makes sense! I'm actually using asyncio-mqtt inside a web framework, that's why I am struggling. Using FastAPI's dependency injection to share context is pretty neat, for other frameworks it sadly doesn't work that easily.

I have some recent (though not direct) experience with FastAPI and it looks like a great framework. Can't say for other frameworks but I fear for the current state of affairs.

---

> I read in e.g. #117 that you're leaning against using `connect/disconnect` directly. I think it could be a nice solution for other web frameworks, usually, there's some sort of `startup/shutdown` procedure. It doesn't seem too unusual to have both either, see [`encode/databases`](https://www.encode.io/databases/connections_and_transactions/).

I'm probably just being stubborn/pedantic about it. 😅 

**In a perfect world**, every (web)framework uses context managers and doesn't even expose raw "on enter"/"on exit" callbacks. There are no footguns since you can *only* use the client this way.
Starlette (that FastAPI builds on) actually supports context managers (https://www.starlette.io/events/) but FastAPI doesn't expose that API yet (see https://github.com/tiangolo/fastapi/pull/2944)

**In the current world**, few-to-none (web)frameworks use context managers and instead expose raw "on enter"/"on exit" callbacks. So we have footguns since our APIs have to expose separate "on enter"/"on exit" functions *and we can't restrict their use*.
Example of misuse: Call `disconnect` in one task while another task still relies on the connection. Or, call `connect` repeatedly in a background task for a crude-but-well-intentioned workaround to auto-reconnect logic.
To support all possible combinations of `connect`/`disconnect`, we have to add extra runtime checks to *all* functions to avoid misuse. This is what we currently (try) to do and it's to blame for a lot of the internal complexity in asyncio_mqtt.

**In any case**, I like to think that I can be pragmatic so in the spirit of pragmatism, let's design an API for the **current world** (and not the **perfect world**). That puts the strain on us (the asyncio-mqtt developers) to ensure that we have the proper runtime checks for `connect`/`disconnect`.

> Automating reconnection would be pretty cool, I already considered that as one of the next things to do 👍

Awesome! I look forward to the discussions/PRs. 😄

frederikaalund:
I'll try to get back on track with this PR now. 😅 

> I now also added an example of how to start a listener inside a web framework without blocking the rest of the code. I'm happy about all feedback! 😊

Looks really good! Well done. 👍 

In practice, I know that starlette depends on `anyio` so if it was me, I'd probably use something like:

```python
@contextlib.asynccontextmanager
async def lifespan(app):
    async with anyio.create_task_group() as tg:
        tg.start_soon(listen)  # This is your `listen` method from the "Side by side with async web frameworks" example
        yield

routes = [
    ...
]

app = Starlette(routes=routes, lifespan=lifespan)
```

To be clear: That's not a "request for change". It's just an alternative example. :) You decide. 👍 

---

As for this PR as a whole it LGTM. It's a great improvement to our documentation and it'll be a welcome resource for new and existing users alike. :)

So if you're happy with it as well, I'll merge right away. **Just give your final confirmation**.

empicano:
Wow, thank you for the thoughtful review, I learned a lot!

---

> You could also choose to do a `Message.matches` [...]

This does look even cleaner, great idea.

---

> Starlette (that FastAPI builds on) actually supports context managers

I didn't know that (or that it's possible like that). Consider my point moot then. I admit I was mostly pushing this for my use case with starlette. The FastAPI PR that you linked looks straightforward, so when that is merged we have an elegant solution for most popular async frameworks, contrary to what I originally thought.

I don't want to storm in and add lots of functionality that later only leads to headaches, either, you're right about wanting to keep it lean. Thanks for the "example of misuse" as well. I'm onboard now with the rejection of `connect`/`disconnect` 😄

---

Last changes:

- I included the starlette example in the "Sharing a connection" section (using asyncio, I think it's nicer to use the stdlib in the examples, even though I'm going to use anyio for my use case now) and rewrote the "Listening without blocking" one to be without starlette
- I added the discussion about exposing `connect`/`disconnect`, hopefully there are going to be less questions on it now, I saw in other issues that you had to defend this quite often already 😉

If you're ok will all that, I think it's ready to merge! 😊

PR #132, opened 2022-10-07
Update related projects list

empicano:
`hbmqtt` is no longer maintained, there exists an active fork called `aqmtt`. I switched those two out and tried to simplify the list of alternatives by generalizing to "Related projects" not just `asyncio`-based. I removed `aiomqtt` as they recommend `asyncio-mqtt` themselves.

Let me know what you think 😊

-- Felix

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,42 +223,19 @@ async with Client(
     ...
 ```
 
-## Alternative asyncio-based MQTT clients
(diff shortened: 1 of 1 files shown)


frederikaalund:
LGTM. Thanks for your contribution to asyncio-mqtt. 👍

PR #133, opened 2022-10-11
Improve the type hints

JonathanPlasse:
Hi,
I formatted the code with black and isort. I could roll it back if you want.
I use [types-paho-mqtt](https://github.com/python/typeshed/pull/8878) to update the type hints.
It pass mypy strict mode.


asyncio_mqtt/client.py
```diff
@@ -312,20 +327,37 @@ async def force_disconnect(self) -> None:
...
+            str,
+            tuple[str, mqtt.SubscribeOptions],
+            list[tuple[str, mqtt.SubscribeOptions]],
+            list[tuple[str, int]],
+        ],
+        qos: int = 0,
+        options: Optional[mqtt.SubscribeOptions] = None,
+        properties: Optional[Properties] = None,
+        timeout: int = 10,
+    ) -> int:
```
frederikaalund, line 342: I like the idea to explicitly type these parameters. Makes it easier to use the library.

I know that you exhaust the current argument list for `paho.Client.subscribe`. In any case, let's keep the `*args` and `**kwargs` forwarding just for the sake of future compatibility. 👍 This way, if paho adds a new argument, asyncio-mqtt is still compatible (even though we don't expose explicit typings for ssaid argument).

Of course, this goes for the other functions as well.

asyncio_mqtt/client.py
```diff
@@ -490,7 +530,7 @@ def _on_connect(
         client: mqtt.Client,
         userdata: Any,
         flags: Dict[str, int],
-        rc: int,
+        rc: int | mqtt.ReasonCodes,
```
frederikaalund, line 533: Nice touch of Python 3.10. :)) For the sake of backward-compatibility, let's use `Union` for now. With time, we can get rid of this legacy stuff (but not now).

asyncio_mqtt/client.py
```diff
@@ -555,7 +593,7 @@ def _on_unsubscribe(
         userdata: Any,
         mid: int,
         properties: Optional[mqtt.Properties] = None,
-        reasonCodes: Optional[List[mqtt.ReasonCodes]] = None,
+        reasonCodes: Optional[List[mqtt.ReasonCodes] | mqtt.ReasonCodes] = None,
```
frederikaalund, line 596: See above (about `Union` over `|` for now).

asyncio_mqtt/client.py
```diff
@@ -92,37 +97,38 @@ def __init__(
         certfile: Optional[str] = None,
         keyfile: Optional[str] = None,
         cert_reqs: Optional[ssl.VerifyMode] = None,
-        tls_version: Optional[ssl._SSLMethod]= None,
+        tls_version: Optional[ssl._SSLMethod] = None,
         ciphers: Optional[str] = None,
-        keyfile_password: Optional[str] = None       
+        keyfile_password: Optional[ssl._PasswordType] = None,
```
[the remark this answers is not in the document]
frederikaalund: Same goes for `ssl._SSLMethod` above.

asyncio_mqtt/client.py
```diff
@@ -182,20 +188,22 @@ def __init__(
...
-        self._connected: "asyncio.Future[int]" = asyncio.Future()
-        self._disconnected: "asyncio.Future[Optional[int]]" = asyncio.Future()
+        self._connected: "asyncio.Future[Union[int, mqtt.ReasonCodes]]" = (
+            asyncio.Future()
+        )
+        self._disconnected: "asyncio.Future[Union[int, mqtt.ReasonCodes, None]]" = (
+            asyncio.Future()
+        )
         # Pending subscribe, unsubscribe, and publish calls
-        self._pending_subscribes: Dict[int, "asyncio.Future[int]"] = {}
```
frederikaalund, line 188: See later comments about this change. :)

asyncio_mqtt/client.py
```diff
@@ -230,31 +238,32 @@ def __init__(
...
+
         if proxy is not None:
-            self._client.proxy_set(**proxy.proxy_args)    
+            self._client.proxy_set(**proxy.proxy_args)
 
-        if websocket_path is not None or websocket_headers is not None:
-            self._client.ws_set_options(
-                path=websocket_path, headers=websocket_headers
-            )
+        if websocket_path is not None:
```
frederikaalund, line 256: Good catch here. The `if` before was erroneous.

asyncio_mqtt/client.py
```diff
@@ -230,31 +238,32 @@ def __init__(
...
 
         if will is not None:
             self._client.will_set(
                 will.topic, will.payload, will.qos, will.retain, will.properties
             )
 
         self._client.message_retry_set(message_retry_set)
-        self._socket_options: Tuple[SocketOption] = tuple(socket_options)
+        if socket_options is not None:
+            self._socket_options: Tuple[SocketOption, ...] = tuple(socket_options)
```
frederikaalund, line 267: Instead, change the default of `socket_options` to `None` (in the function parameters).
Then, use:
```python
if socket_options is not None:
    socket_options = tuple()
self._socket_options: Tuple[SocketOption, ...] = tuple(socket_options)
```

frederikaalund: Note that way *always* assign `self._socket_options` now. This is important. :)

asyncio_mqtt/client.py
```diff
@@ -539,13 +579,11 @@ def _on_subscribe(
...
         mid: int,
-        granted_qos: int,
+        granted_qos: Union[Tuple[int], List[mqtt.ReasonCodes]],
         properties: Optional[mqtt.Properties] = None,
     ) -> None:
         try:
-            fut = self._pending_subscribes.pop(mid)
-            if not fut.done():
-                fut.set_result(granted_qos)
+            self._pending_subscribes.pop(mid).set()
```
frederikaalund, line 586: See above. Revert for now. It's not time for a breaking change. Also, the `if` may seem redundant but it actually avoids a race condition. Trust me. :)

frederikaalund (changes requested):
Thanks for this in-depth cleanup of the code! :) See my review for details. Let me know if you have any questions. 👍 

~Frederik

asyncio_mqtt/client.py
```diff
@@ -312,20 +327,37 @@ async def force_disconnect(self) -> None:
...
         # Early out on error
         if result != mqtt.MQTT_ERR_SUCCESS:
             raise MqttCodeError(result, "Could not subscribe to topic")
         # Create future for when the on_subscribe callback is called
-        cb_result: "asyncio.Future[int]" = asyncio.Future()
+        cb_result = asyncio.Event()
         with self._pending_call(mid, cb_result, self._pending_subscribes):
             # Wait for cb_result
-            return await self._wait_for(cb_result, timeout=timeout)
+            return await self._wait_for(cb_result.wait(), timeout=timeout)
```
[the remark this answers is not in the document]
JonathanPlasse: The type annotation was wrong, the `Future` was of type `int` instead of the actual type `Union[Tuple[int], List[mqtt.ReasonCodes]]`.
So, the return should be of type `Union[Tuple[int], List[mqtt.ReasonCodes]]`.
It would be a breaking change to now really return an `int` instead of `Union[Tuple[int], List[mqtt.ReasonCodes]]`.

asyncio_mqtt/client.py
JonathanPlasse: Do you mean
```python
if socket_options is None:
    socket_options = tuple()
self._socket_options: Tuple[SocketOption, ...] = tuple(socket_options)
```

frederikaalund: Yes. 😁 Good catch. 👍

JonathanPlasse:
Should I squash and rebase the commits?

JonathanPlasse:
Change `Awaitable` with `Coroutine` for `_outgoing_call` as returning the former caused a type error with `anyio.abc.TaskGroup.start_soon`.

frederikaalund (approved):
Well done. 👍 I have no further comments. LGTM.

frederikaalund:
> Should I squash and rebase the commits?

I can (and probably will) do that through the GitHub UI so you don't have to. :)

> Change Awaitable with Coroutine for _outgoing_call as returning the former caused a type error with anyio.abc.TaskGroup.start_soon.

Another good catch! Well done. 👍 

---

I'm very happy with this PR. Thank you for your contribution to asyncio-mqtt! Keep it coming. 😄

JonathanPlasse:
Thanks

PR #134, opened 2022-10-15
Fix autocomplete for _outgoing_call decorator

JonathanPlasse:
The signature of the `Client.subscibe()` for example only displayed
`(...) -> Coroutine[Any, Any, Tuple[int] | List[ReasonCodes]]` 
instead of
`(topic: str | Tuple[str, SubscribeOptions] | List[Tuple[str, SubscribeOptions]] | List[Tuple[str, int]], qos: int = ..., options: SubscribeOptions | None = ..., properties: Properties | None = ..., *args: Any, timeout: int = ..., **kwargs: Any) -> Coroutine[Any, Any, Tuple[int] | List[ReasonCodes]]`

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -29,11 +29,15 @@ from typing import (
     cast,
 )
 
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
 if sys.version_info >= (3, 7):
     from contextlib import asynccontextmanager
 else:
     from async_generator import asynccontextmanager as _asynccontextmanager
-    from typing_extensions import ParamSpec
 
     _P = ParamSpec("_P")
     _T = TypeVar("_T")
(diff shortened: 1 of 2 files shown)


frederikaalund:
Thanks for taking a deep look at the typings. 👍 

I like this direction with `ParamSpec` and friends. It's the way to go.

My only comment is that you should conditionally import `ParamSpec`:
 * from `typing` if we are on 3.10 or later.
 * from `typing_extensions` otherwise.

This way, typing-extensions remains an *optional* dependency (and not a requirement) depending on the python version. See https://github.com/sbtinstruments/asyncio-mqtt/blob/master/setup.py#L40 for details.

JonathanPlasse:
Done

JonathanPlasse:
When do you think you will release a new version?

frederikaalund:
> When do you think you will release a new version?

I can do that tomorrow. :)

Thank you for your contribution to asyncio-mqtt.

JonathanPlasse:
You are welcome!

frederikaalund:
v0.13.0 is out now. 👍

PR #140, opened 2022-10-27
docs: reorder README sections in order of importance

empicano:
Hi Frederik!

Again with the README 😄 I tweaked a few little things, and reordered the sections, among others to move the _Note for Windows users_ higher, as there are quite a few people who seem to overlook this. (Btw. is there anything that speaks against implementing the switch to the `SelectorEventLoop` directly in the code? I guess changing the loop silently is not a very nice pattern.)

~ Felix

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
-![license](https://img.shields.io/github/license/sbtinstruments/asyncio-mqtt)
-![semver](https://img.shields.io/github/v/tag/sbtinstruments/asyncio-mqtt?sort=semver)
-[![PyPI](https://img.shields.io/pypi/v/asyncio-mqtt)](https://pypi.org/project/asyncio-mqtt/)
-
-# Idiomatic asyncio MQTT client 🙌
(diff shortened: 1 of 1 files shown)


JonathanPlasse:
Why delete the requirement section?

empicano:
I didn't delete it, I just merged it with the installation section 😉

frederikaalund:
Good idea to move the "Note for Windows users" section up. 👍 We get a frequent stream of new issues due to that.

> (Btw. is there anything that speaks against implementing the switch to the SelectorEventLoop directly in the code? I guess changing the loop silently is not a very nice pattern.)

Technically, there is no barrier. :) It's just intrusive to switch the loop like that IMHO. What we could do, is to issue a warning if the user attempts to use the library with a non-supported event loop. In that warning, we can instruct the user to switch the event loop. I weldcome a PR with that change. 👍 

Back to this PR itself. LGTM. Thanks!

~Frederik

PR #146, opened 2022-10-30
Drop Python 3.6

### LINKED ISSUE #141: Should Python 3.6 support be dropped?
Python 3.6 reached its end of life nearly a year ago.
So, maybe we should remove its support.

JonathanPlasse:
Close #141

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
+from __future__ import annotations
+
 import asyncio
 import functools
 import logging
@@ -6,13 +8,12 @@ import socket
 import ssl
 import sys
 from contextlib import contextmanager
+from dataclasses import dataclass
 from enum import IntEnum
 from types import TracebackType
 from typing import (
     Any,
-    AsyncContextManager,
     AsyncGenerator,
-    AsyncIterator,
     Awaitable,
     Callable,
     Coroutine,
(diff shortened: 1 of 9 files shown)


asyncio_mqtt/client.py
```diff
@@ -1,52 +1,37 @@
...
 from types import TracebackType
 from typing import (
     Any,
-    AsyncContextManager,
     AsyncGenerator,
-    AsyncIterator,
     Awaitable,
     Callable,
     Coroutine,
     Dict,
```
empicano, line 20: I'm not very experienced with typing in Python. Is this on purpose that you're leaving some `typing.Dict`, `typing.Tuple`, and `typing.Union` in the code, instead of changing everything to `dict`, `tuple` and `|`?

JonathanPlasse: Yes, the typing used as type annotations can be `dict`, `tuple`, and `|` as they are not executed with `from __future__ import annotations`, but types that are assigned to variables are executed so they must remain as they are.

empicano: And the non-`typing` versions `dict`, `tuple`, and `|` mustn't be executed because they are not valid python<3.10 code, is that it?

Neat 😎

JonathanPlasse: Yes

frederikaalund:
LGTM. :+1:

I like the use of the new-style `dict`, `type`, `list`, `|` typings combined with `from __future__ import annotations`. Good way to use new features in a backwards-compatible way.

In a future commit, let's add some type aliases for convenience. E.g.:
```python
Topic = (
    str
    | tuple[str, mqtt.SubscribeOptions]
    | list[tuple[str, mqtt.SubscribeOptions]]
    | list[tuple[str, int]]
)
```

Even if we only use the alias in a single place, it's a little verbose to read through as part of a function parameter list.

In any case, good job with this PR. :+1: Great to see that you also fixed the tests and added the "supported python versions" badge to the documentation (and setup).

frederikaalund (approved):
Approved. :)

I see that you still some recent commits in this PR/branch. Do you want to add some other finishing touches before the merge or can I go ahead and merge it as-is?

JonathanPlasse:
> LGTM. +1
> 
> I like the use of the new-style `dict`, `type`, `list`, `|` typings combined with `from __future__ import annotations`. Good way to use new features in a backwards-compatible way.
> 
> In a future commit, let's add some type aliases for convenience. E.g.:
> 
> ```python
> Topic = (
>     str
>     | tuple[str, mqtt.SubscribeOptions]
>     | list[tuple[str, mqtt.SubscribeOptions]]
>     | list[tuple[str, int]]
> )
> ```
> 
> Even if we only use the alias in a single place, it's a little verbose to read through as part of a function parameter list.
> 
> In any case, good job with this PR. +1 Great to see that you also fixed the tests and added the "supported python versions" badge to the documentation (and setup).

The user would lose the details and only see the type alias. Is it what we want.

JonathanPlasse:
It is ready to be merged.
Which options do you use to merge?
- Create a merge commit
- Rebase and merge

For me, the cleaner option for a better history is rebasing.

frederikaalund:
*Rebase and merge*: If the commits are "clean" (e.g., not "TODO: Fix this" or "WIP: still developing on X").
*Squash and merge*: Otherwise.
*Create a merge commit*: Never.

For this PR, I think that we are in the first category: *Rebase and merge*. :+1:

JonathanPlasse:
I removed the option *Create a merge commit* to avoid errors.

PR #149, opened 2022-11-01
Add more tests

JonathanPlasse:


JonathanPlasse:
I added tests for nearly any configuration of `Client`. The rest of the missing coverage is error handling.
@frederikaalund, @empicano, any feedback.

frederikaalund:
I really like this initiative to add more "robustness" to asyncio-mqtt via tests. 👍 Great job on that. 😃 
It looks good from my cursory reading!

I don't have enough time tonight to give the proper review that this effort deserves (I already wrote a lengthy reply to another PR). I'll do a review later this weekend when I have the time. :)

Sorry about the lack of time on my part. It's been a busy week at work. I'll try to make up for it this weekend.


~Frederik

asyncio_mqtt/client.py
```diff
@@ -300,12 +308,7 @@ async def force_disconnect(self) -> None:
     @_outgoing_call
     async def subscribe(
         self,
-        topic: (
-            str
-            | tuple[str, mqtt.SubscribeOptions]
-            | list[tuple[str, mqtt.SubscribeOptions]]
-            | list[tuple[str, int]]
-        ),
+        topic: SubscribeTopic,
```
frederikaalund, line 311: Good to see this compelx type definition moved out of the function definition. 👍

asyncio_mqtt/error.py
```diff
@@ -31,10 +31,8 @@ def __init__(self, rc: int | mqtt.ReasonCodes):
         if isinstance(rc, mqtt.ReasonCodes):
             return super().__init__(rc)
         msg = "Connection refused"
-        try:
+        if rc in _CONNECT_RC_STRINGS:
             msg += f": {_CONNECT_RC_STRINGS[rc]}"
-        except KeyError:
-            pass
         super().__init__(rc, msg)
```
frederikaalund, line 36: The pythonic way is to ask for forgiveness, not permission. In other languages (like C++), it's the other way around.
In any case, let's keep the (admittedly verbose) `try..except` block to stay pythonic.

See https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain

tests/conftest.py
```diff
@@ -1,16 +1,13 @@
...
 @pytest.fixture
-def os_and_version() -> str:
-    return sys.platform + "_" + ".".join(map(str, sys.version_info[:2]))
-
-
-@pytest.fixture
-def anyio_backend() -> Tuple[str, Dict[str, Any]]:
+def anyio_backend() -> tuple[str, dict[str, Any]]:
     if sys.platform == "win32":
         from asyncio.windows_events import WindowsSelectorEventLoopPolicy
```
frederikaalund, line 12: Maybe we should do something similar to this in the regular lib as well. We get a lot of issues from Windows users due to this.

tests/test_client.py
```diff
@@ -52,25 +64,25 @@ async def handle_filtered_messages() -> None:
             await client.publish(topic_unfiltered, 2)
 
 
-async def test_client_unsubscribe(os_and_version: str) -> None:
-    topic_header = os_and_version + "/tests/asyncio_mqtt/unsubscribe/"
+async def test_client_unsubscribe() -> None:
+    topic_header = TOPIC_HEADER + "unsubscribe/"
     topic1 = topic_header + "1"
     topic2 = topic_header + "2"
     event = anyio.Event()
```
frederikaalund, line 71: Maybe I'm blind but what's the purpose of this `event`? We set it but never await it.

tests/test_client.py
```diff
@@ -108,5 +120,157 @@ async def launch_client() -> None:
...
+    async def handle_messages(tg: anyio.abc.TaskGroup) -> None:
+        async with client.filtered_messages(topic) as messages:
+            async for message in messages:
+                assert message.topic == topic
+                tg.cancel_scope.cancel()
+
+    async with Client(
+        HOSTNAME,
+        8883,
+        tls_context=ssl.SSLContext(protocol=ssl.PROTOCOL_TLS),
```
frederikaalund, line 139: Good to see all these "optional parameter" code paths exercised! 👍

tests/test_client.py
```diff
@@ -108,5 +120,157 @@ async def launch_client() -> None:
...
+        async with anyio.create_task_group() as tg:
+            await client.subscribe(topic)
+            tg.start_soon(handle_messages, tg)
+            await client.publish(topic)
+
+
+async def test_client_logger() -> None:
+    logger = logging.getLogger("asyncio-mqtt")
+    async with Client(HOSTNAME, logger=logger) as client:
+        assert logger == client._client._logger  # type: ignore[attr-defined]
```
frederikaalund, line 188: Prefer `is` for these kind of "exactly this object" comparisons. :)

tests/test_client.py
```diff
@@ -108,5 +120,157 @@ async def launch_client() -> None:
...
+
+
+async def test_client_max_concurrent_outgoing_calls(
```
frederikaalund, line 191: Now this is an interesting test! Well done. 👍

I wonder how many people actually use this feature in production. 😄 

This also showcases a weakness in our current API: We don't have any backpressure mechanics. *That* is difficult to solve "right", though. Definitely a 2.0 kind of thing (or with the anyio rewrite/replacement).

If you're interested in that kind of stuff there is a good discussion here: https://github.com/python-trio/trio/issues/987

tests/test_client.py
```diff
@@ -108,5 +120,157 @@ async def launch_client() -> None:
...
+
+    monkeypatch.setattr(mqtt, "Client", MockPahoClient)
+
+    async with Client(HOSTNAME, max_concurrent_outgoing_calls=1) as client:
+        await client.subscribe(topic)
+        await client.unsubscribe(topic)
+        await client.publish(topic)
+
+
+async def test_client_websockets() -> None:
```
frederikaalund, line 238: Really good to cover ground here. 👍 I know that many of our users use MQTT piped through a WebSocket connection.

tests/test_client.py
```diff
@@ -108,5 +120,157 @@ async def launch_client() -> None:
...
+        websocket_path="/",
+        websocket_headers={"foo": "bar"},
+    ) as client:
+        async with anyio.create_task_group() as tg:
+            await client.subscribe(topic)
+            tg.start_soon(handle_messages, tg)
+            await client.publish(topic)
+
+
+async def test_client_pending_calls_threshold(caplog: pytest.LogCaptureFixture) -> None:
```
frederikaalund, line 260: That's a nice fixture. Didn't know about it. :)

tests/test_client.py
```diff
@@ -108,5 +120,157 @@ async def launch_client() -> None:
...
+
+        async with anyio.create_task_group() as tg:
+            for _ in range(nb_publish):
+                tg.start_soon(client.publish, topic)
+
+        assert caplog.record_tuples == [
+            (
+                "mqtt",
+                logging.WARNING,
+                f"There are {nb_publish} pending publish calls.",
```
frederikaalund, line 274: See my earlier comment about backpressure. The current "just log it and continue" approach of mine is *very* crude. 😅

In any case, good to have the behaviour covered in tests. 👍

tests/test_error.py
```diff
@@ -0,0 +1,72 @@
...
+        mqtt.MQTT_ERR_UNKNOWN,
+        mqtt.MQTT_ERR_ERRNO,
+        mqtt.MQTT_ERR_QUEUE_SIZE,
+        mqtt.MQTT_ERR_KEEPALIVE,
+        -1,
+    ),
+)
+def test_mqtt_code_error_int(rc: int) -> None:
+    assert str(MqttCodeError(rc)) == f"[code:{rc}] {mqtt.error_string(rc)}"
+
```
frederikaalund, line 33: Nice use of test parameterization. 👍 

These tests are a bit strict IMHO. I wouldn't call the return value of `MqttCodeError.__str__` as part of the stable API. I fear that the presence of this test make people assume that this *is* indeed part of the stable API. In any case, we don't have any plans to change it either, so I guess why not. :)

frederikaalund (approved):
LGTM. 👍 Well done! 😄

I put some minor remarks in the comments (mostly pointing to future API decisions). I don't have any requests for changes.

Really great to see our test coverage increase like this. I especially like how you exercised some of the subtle details in our API (the "backpressure" semantics). I think that most people don't realize how important that aspect actually is. Not before it's too late at least! 😄 See some of the early issues for reference.

asyncio_mqtt/error.py
frederikaalund: It's a bit nitpicky on my side, I'll admit that. 😄 You decide which version that you want to keep here.

tests/test_client.py
JonathanPlasse: Is not `max_concurrent_outgoing_calls` a backpressure mechanism?

frederikaalund: It is but it is a very crude one at that. 😄 My point is that we can and should do better in a future API.

JonathanPlasse:
I added a test that there is no warning for pending calls when `max_concurrent_outgoing_calls` is set.

PR #152, opened 2022-11-12
AnyIO rework

JonathanPlasse:
BREAKING CHANGE: This breaks support for the old asyncio-based API.
The client behaves differently based on if it runs on `localhost` or `test.mosquitto.org`.
The test suite does not pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ local_test.py
 .coverage
 coverage.xml
 asyncio_mqtt/_version.py
+.vscode/
--- a/README.md
+++ b/README.md
@@ -1,4 +1,11 @@
-<h1 align="center">Idiomatic asyncio MQTT Client 🙌</h1>
+# This is a re-write based on anyio
+
+Maybe we should fork this to another repo.
+
+I added `anyio_example.py` that you can test with locally. When we agree
+on the API, we can update the rest of this readme.
+
+<h1 align="center">Idiomatic AnyIO MQTT Client 🙌</h1>
 <p align="center">
     <a href="https://github.com/sbtinstruments/asyncio-mqtt/blob/main/LICENSE"><img alt="License: BSD-3-Clause" src="https://img.shields.io/github/license/sbtinstruments/asyncio-mqtt"></a>
     <a href="https://pypi.org/project/asyncio-mqtt"><img alt="PyPI version" src="https://img.shields.io/pypi/v/asyncio-mqtt"></a>
--- a/anyio_example.py
+++ b/anyio_example.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+
+from contextlib import _AsyncGeneratorContextManager
+from random import randrange
+from typing import AsyncGenerator
+
+import anyio
+from paho.mqtt.client import MQTTMessage
+
+from asyncio_mqtt import Client, MqttError
+
+
+async def advanced_example(client: Client) -> None:
+    async with anyio.create_task_group() as tg:
+        # You can create any number of topic filters
+        topic_filters = (
+            "floors/+/humidity",
+            "floors/rooftop/#"
+            # 👉 Try to add more filters!
+        )
+        for topic_filter in topic_filters:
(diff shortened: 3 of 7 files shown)


frederikaalund:
Hi Jonathan, thanks for opening this PR. 👍 Let me have a look. :)

Ah, I remember this. It's the work from the `feat-anyio` branch, right? Good job with the overall modernization/rebasing of said branch. 👍 

When I made this branch back then, I didn't move forward for the following reasons:

 1. This is a breaking change that I don't think most `asyncio` users want.
 2. It adds another dependency (`anyio` that in turn has one or two dependencies).
 3. I think the `Client` API itself could use an overhaul if we do a breaking change anyhow.
 4. Implementation-wise, I made the `Promise` class to shoehorn the anyio-based internals into the `Client` API. It felt wrong back then and it feels wrong today. There must be a better way (probably leading to a better API). On that note, I think that [@flyte's implementation is interesting](https://github.com/flyte/anyio-mqtt/blob/develop/anyio_mqtt/__init__.py).

Points (1) and (2) alone were the deciding factors for me to not move forward back then.

All that being said, I do think that a anyio-based MQTT client is the way forward. The big question for is whether we do this as:

 1. Version "2.0" of asyncio-mqtt. Maybe rebrand to aiomqtt (I have the author's permission to inherit this name)
 2. A completely new anyio-based project (e.g., anyio-mqtt)
 3. A completely new [sans-IO](https://sans-io.readthedocs.io/)-based project with an anyio-based "back-end"

Note that point (3) requires a re-write of the MQTT protocol itself, which defauts the original purpose of "use paho-mqtt for it's battle-tested stability".

What do you (asyncio-mqtt maintainers/community) think? :)

empicano:
> 1. Version "2.0" of asyncio-mqtt. Maybe rebrand to aiomqtt (I have the author's permission to inherit this name)

If we stay scoped to AsyncIO, independent of asyncio vs. anyio, I'm all for this 👍 

> 3. A completely new [sans-IO](https://sans-io.readthedocs.io/)-based project with an anyio-based "back-end"

The concept sounds interesting. With unlimited maintainer resources, this would probably be the best way forward. paho-mqtt has not seen much love from eclipse for a year now, either, I don't know of any official statement though.

I didn't know of sans-IO before, I found these discussions concerning uvicorn and [hypercorn](https://github.com/pgjones/hypercorn) interesting:
- https://github.com/encode/uvicorn/issues/47
- https://github.com/encode/uvicorn/discussions/1105

Both use hyper's h11 library, hypercorn also the h2 one.

frederikaalund:
Lots of good discussions in the linked posts. Thanks for sharing! 👍 

We face very similar issues here.

empicano:
On re-reading [the discussion](https://github.com/sbtinstruments/asyncio-mqtt/discussions/44) about this and your thoughts, Frederik, I particularly agree with your first point: "This is a breaking change that I don't think most `asyncio` users want."

I think this project started as `asyncio` and people seem to like that, so we shouldn't change that.

Three ideas I'd be on board with:

- Looking into how to integrate the new `asyncio` structured concurrency features of Python 3.11 into asyncio-mqtt (our code, documentation, etc.)
- A rebrand to aiomqtt
- A switch from paho-mqtt to a sans-io "backend". We would need to write this ourselves in an extra project, of course, but if I understand this right, we could leave our API mostly as is. With that, we could remove some paho-mqtt quirks, and essentially switch our only dependency against a more lightweight one.

We can still build a separate `anyio`-based project, but the best way forward with this would probably be by continuing [flyte's project](https://github.com/flyte/anyio-mqtt).

JonathanPlasse:
I have the project to write a wrapper of [mqttbytes](https://crates.io/crates/mqttbytes) with PyO3.
It would be a Sans IO library.

empicano:
> I have the project to write a wrapper of [mqttbytes](https://crates.io/crates/mqttbytes) with PyO3.

That sounds pretty cool!!

JonathanPlasse:
I will start working on it during the holidays.

PR #155, opened 2022-11-16
Replace flake8 by ruff 🚀

JonathanPlasse:
[Ruff](https://github.com/charliermarsh/ruff) is the new `flake8` (near instant linting).
Check out the [launch blog article](https://notes.crmarsh.com/python-tooling-could-be-much-much-faster).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-ignore = E301, E302, E501, W503
-per-file-ignores = setup.py: F821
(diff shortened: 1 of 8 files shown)


frederikaalund:
I'm all for this. I've followed `ruff` for a while and it looks to be the next big thing for static analysis in python. :)

Personally, I'd use `ruff` alongside `mypy`, `pylint`, and `black`. That's because I care more about "stati analysis coverage" than speed. It's a tradeoff between development velocity and code quality guarantees.

In any case, I vote that we go ahead and replace `flake8` with `ruff`. 👍 There is no contest there. 😄

PR #170, opened 2022-12-22
Don't call add_reader/writer from executor thread

mmmspatz:
It isn't safe to call `loop.add_reader()` and `loop.add_writer()` from a different OS thread: https://peps.python.org/pep-3156/#thread-interaction

Because we run `mqtt.Client.connect()` in an executor, it's possible for both `asyncio-mqtt.Client._on_socket_open()` and `asyncio-mqtt.Client._on_socket_register_write()` to run outside the main event loop thread (these are callbacks that are invoked by `mqtt.Client.connect()`)

So, follow the example that was already in place for creating the task that runs `self._misc_loop()` to call `add_reader()` and `add_writer()` in a thread-safe manner as well.

I noticed this issue when using the event loop provided by [qasync](https://github.com/CabbageDevelopment/qasync). My client always timed out while connecting, because the reader and writer weren't functioning. With this change, it works as expected.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -775,14 +775,16 @@ class Client:
                 if not self._disconnected.done():
                     self._disconnected.set_exception(exc)
 
-        self._loop.add_reader(sock.fileno(), callback)
         # paho-mqtt calls this function from the executor thread on which we've called
-        # `self._client.connect()` (see [3]), so we create a callback function to
-        # schedule `_misc_loop()` and run it on the loop thread-safely.
-        def create_task_callback() -> None:
+        # `self._client.connect()` (see [3]), so we can't do most operations on
+        # self._loop directly. Instead, wrap them in a coroutine and run that
+        # thread-safely.
+        async def on_loop_coroutine() -> None:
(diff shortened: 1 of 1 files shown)


frederikaalund:
Hi Mark, thanks for creating this PR. Let me have a look.

You're right, we need to move all the calls to `_loop´ methods into the thread-safe context. That makes total sense and good job with both the discovery and fix of this bug.

I'm a little concerned about the `fut.result()` line since it'll block until completion. I guess that this is okay since the entire `_on_socket_open`/`_on_socket_close` callbacks run inside another thread in the first place (in turn, this is why we have to make the calls inside thread-safe).

> I noticed this issue when using the event loop provided by [qasync](https://github.com/CabbageDevelopment/qasync). My client always timed out while connecting, because the reader and writer weren't functioning. With this change, it works as expected.

I also used the qasync event loop together with asyncio-mqtt! I never ran into this issue. Probably because I'm on a *very* old version of asyncio-mqtt in our production code (don't change it if it ain't broken).

In any case, nice catch and good fix. I'm just waiting until the CI checks complete before I merge. 👍

PR #175, opened 2023-01-04
fix: Don't block on add_reader/writer future

### LINKED ISSUE #174: Client._on_socket_register_write() may deadlock
Oops, I introduced a bug in #170. Sorry.

`Client._on_socket_register_write()` is indeed called from an executor thread at startup, and needs to interact with the main event loop in a thread-safe matter. But later, it is also called from within the event loop (I believe regularly as data is sent on the socket). 

When called from `self._loop`, the following code will deadlock:
``` python
 fut = asyncio.run_coroutine_threadsafe(on_loop_coroutine(), self._loop)
 fut.result()
```

Waiting on `fut` prevents control from being returned to the event loop, so `on_loop_coroutine()` never runs, so `fut` never becomes ready.

mmmspatz:
Client._on_socket_register_writer() may be called from the event loop, in which case adding a Task to that loop then blocking on its completion will deadlock.

Fixes #174

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -777,14 +777,12 @@ class Client:
 
         # paho-mqtt calls this function from the executor thread on which we've called
         # `self._client.connect()` (see [3]), so we can't do most operations on
-        # self._loop directly. Instead, wrap them in a coroutine and run that
-        # thread-safely.
-        async def on_loop_coroutine() -> None:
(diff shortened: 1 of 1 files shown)


frederikaalund:
Thank you for the quick follow up and patch. :+1: LGTM!

Sorry for the slow response form my side. I've been on vacation.

~Frederik

PR #176, opened 2023-01-05
fix: consistently use a user-provided logger

rnovatorov:
Client initializer currently accepts a logger as an optional argument. That logger is passed to the underlying paho.mqtt library and then used for logging there. However, there also exists MQTT_LOGGER, which is used internally by asyncio-mqtt. Without looking at the source code it is easy to forget that in order to enable logging both in paho.mqtt and in asyncio-mqtt it is not enough to pass a logger as an argument to Client initializer.

This PR makes asyncio-mqtt use a user-provided logger instead of MQTT_LOGGER for internal logging, so that if user decides to pass a logger to Client constructor, that logger will be used both in paho.mqtt and asyncio-mqtt.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -317,6 +317,7 @@ class Client:
 
         if logger is None:
             logger = MQTT_LOGGER
+        self._logger = logger
         self._client.enable_logger(logger)
 
         if username is not None:
@@ -485,7 +486,7 @@ class Client:
         self, topic_filter: str, *, queue_maxsize: int = 0
     ) -> AsyncGenerator[AsyncGenerator[mqtt.MQTTMessage, None], None]:
         """Return async generator of messages that match the given filter."""
-        MQTT_LOGGER.warning(
+        self._logger.warning(
             "filtered_messages() is deprecated and will be removed in a future version."
             " Use messages() together with Topic.matches() instead."
         )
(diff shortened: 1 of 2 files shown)


asyncio_mqtt/client.py
```diff
@@ -233,7 +233,7 @@ def __init__(  # noqa: C901
         *,
         username: str | None = None,
         password: str | None = None,
-        logger: logging.Logger | None = None,
+        logger: logging.Logger = MQTT_LOGGER,
```
frederikaalund, line 236: Prefer to use `None` for default values. It makes it easier to extend `Client` (e.g., through inheritance).

In turn, assign the actual default value inside the function itself. E.g.:

```python
if logger is None:
    logger = MQTT_LOGGER
self._logger = logger
self._client.enable_logger(logger)
```

asyncio_mqtt/client.py
```diff
@@ -315,8 +315,7 @@ def __init__(  # noqa: C901
         if max_queued_messages is not None:
             self._client.max_queued_messages_set(max_queued_messages)
 
-        if logger is None:
-            logger = MQTT_LOGGER
+        self._logger = logger
```
frederikaalund, line 318: See previous comment.

frederikaalund (changes requested):
Sorry for the long delay. I must have missed this PR to begin with. Thanks for the ping, @JonathanPlasse. Let me have a look! :)

This looks good to me. :+1:  Well done! I only have one gripe regarding default arguments. See the review for details.

With that small change, I'm ready to merge the code.

~Frederik

asyncio_mqtt/client.py
rnovatorov: Done.

frederikaalund (approved):

frederikaalund:
LGTM. :+1: 

Thank you for your contribution to asyncio-mqtt. :)

I apologize for the unnecessarily long PR review time on my part. I completely missed this PR (must have cleared my inbox a bit too fast :sweat_smile: ).

rnovatorov:
That's not a problem at all. Thank you for the project :)

PR #179, opened 2023-01-10
add(feat): to constructor

AndreasHeine:


frederikaalund:
Hi Andreas, thanks for creating this PR. Let me have a look. :)

Looks good! :+1:  I only have one comment: Use `None` as the default value (to enable easier default chaining). In turn, use `if val is None: val = "default val"` inside the `__init__` body. Does it make sense?

AndreasHeine:
do you mean:

```python
        max_inflight_messages: int | None = 20,
        max_queued_messages: int | None = 0

### ...

        if max_inflight_messages is not None:
            self._client.max_inflight_messages_set(max_inflight_messages)
        if max_queued_messages is not None:
            self._client.max_queued_messages_set(max_queued_messages)

```

frederikaalund:
Exactly, great job! :)

Last thing: Give yourself contribution credit in `CHANGELOG.md`. After that, I'll merge the PR right away. Thanks again!

frederikaalund:
LGTM!

I'd like to merge this right away but there are merge conflicts.

Two options from here:
  1. Use `git rebase -i` (or similar) to resolve the conflicts. I think the two commits from the pre-commit bot accidentally found their way into this PR.
  2. Create a new PR with the same changes but a "clean" commit history without conflicts.

Does it make sense? :)

AndreasHeine:
#180

PR #192, opened 2023-02-12
Make default timeout configurable

### LINKED ISSUE #116: Connection Timeout with Context Manager
The `connect` method does provide a `timeout` argument. But the `__aenter__` method doesn't. So connecting to an unnexisting MQTT broker will make the context manager hang.  
A simple solution would be to be able to create the `Client` object with a default timeout to be used in the `__aenter__` method.

What do you think ?

skewty:
fixes #116

NOTE: `timeout` field of `Client._wait_for` has type `float` but default value all over code is `int`.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -250,7 +250,7 @@ class Client:
         will: Will | None = None,
         clean_session: bool | None = None,
         transport: str = "tcp",
-        timeout: int | float = 10,
+        timeout: float | None = None,
         keepalive: int = 60,
         bind_address: str = "",
         bind_port: int = 0,
@@ -361,6 +361,9 @@ class Client:
         if socket_options is None:
             socket_options = ()
         self._socket_options = tuple(socket_options)
+
+        if timeout is None:
+            timeout = 10
         self.timeout = timeout
 
     @property
@@ -382,7 +385,7 @@ class Client:
         yield from self._pending_unsubscribes.keys()
         yield from self._pending_publishes.keys()
 
-    async def connect(self, *, timeout: int | float | None = None) -> None:
+    async def connect(self, *, timeout: float | None = None) -> None:
         try:
             loop = asyncio.get_running_loop()
 
@@ -406,19 +409,15 @@ class Client:
         # See: https://github.com/eclipse/paho.mqtt.python/blob/v1.5.0/src/paho/mqtt/client.py#L1770
         except (OSError, mqtt.WebsocketConnectionError) as error:
             raise MqttError(str(error)) from None
-        await self._wait_for(
-            self._connected, timeout=self.timeout if timeout is None else timeout
-        )
+        await self._wait_for(self._connected, timeout=timeout)
 
-    async def disconnect(self, *, timeout: int | float | None = None) -> None:
+    async def disconnect(self, *, timeout: float | None = None) -> None:
         rc = self._client.disconnect()
         # Early out on error
         if rc != mqtt.MQTT_ERR_SUCCESS:
             raise MqttCodeError(rc, "Could not disconnect")
         # Wait for acknowledgement
-        await self._wait_for(
-            self._disconnected, timeout=self.timeout if timeout is None else timeout
-        )
+        await self._wait_for(self._disconnected, timeout=timeout)
 
     async def force_disconnect(self) -> None:
         if not self._disconnected.done():
@@ -432,7 +431,7 @@ class Client:
         options: mqtt.SubscribeOptions | None = None,
         properties: Properties | None = None,
         *args: Any,
-        timeout: int | float | None = None,
+        timeout: float | None = None,
         **kwargs: Any,
     ) -> tuple[int] | list[mqtt.ReasonCodes]:
         result, mid = self._client.subscribe(
@@ -447,9 +446,7 @@ class Client:
         ] = asyncio.Future()
         with self._pending_call(mid, callback_result, self._pending_subscribes):
             # Wait for callback_result
-            return await self._wait_for(
-                callback_result, timeout=self.timeout if timeout is None else timeout
-            )
+            return await self._wait_for(callback_result, timeout=timeout)
 
     @_outgoing_call
     async def unsubscribe(
@@ -457,7 +454,7 @@ class Client:
         topic: str | list[str],
         properties: Properties | None = None,
         *args: Any,
-        timeout: int | float | None = None,
+        timeout: float | None = None,
         **kwargs: Any,
     ) -> None:
         result, mid = self._client.unsubscribe(topic, properties, *args, **kwargs)
@@ -468,10 +465,7 @@ class Client:
         confirmation = asyncio.Event()
         with self._pending_call(mid, confirmation, self._pending_unsubscribes):
             # Wait for confirmation
-            await self._wait_for(
-                confirmation.wait(),
-                timeout=self.timeout if timeout is None else timeout,
-            )
+            await self._wait_for(confirmation.wait(), timeout=timeout)
 
     @_outgoing_call
     async def publish(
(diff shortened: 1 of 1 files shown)


JonathanPlasse:
Thanks for your pull-request.
With the [numeric tower](https://peps.python.org/pep-0484/#the-numeric-tower), you only need `float` as it is valid to pass `int` to `float`.

asyncio_mqtt/client.py
```diff
@@ -250,6 +250,7 @@ def __init__(  # noqa: C901
         will: Will | None = None,
         clean_session: bool | None = None,
         transport: str = "tcp",
+        timeout: int | float = 10,
```
frederikaalund, line 253: I propose that we use `timeout: float | None = None`. Note that `int` is redundant as per PEP 484:

> [...] when an argument is annotated as having type float, an argument of type int is acceptable; [...]

Thanks to @JonathanPlasse for pointing that out. +1

Note that we use `None` to mean "Give me the default value". It doesn't mean "no timeout". If you want "no timeout", use `math.inf` instead. :)

asyncio_mqtt/client.py
```diff
@@ -360,6 +361,7 @@ def __init__(  # noqa: C901
         if socket_options is None:
             socket_options = ()
         self._socket_options = tuple(socket_options)
+        self.timeout = timeout
```
frederikaalund, line 364: Replace with:

```python
if timeout is None:
    timeout = 10
self._timeout: float = timeout
```

asyncio_mqtt/client.py
```diff
@@ -380,7 +382,7 @@ def _pending_calls(self) -> Generator[int, None, None]:
         yield from self._pending_unsubscribes.keys()
         yield from self._pending_publishes.keys()
 
-    async def connect(self, *, timeout: int = 10) -> None:
+    async def connect(self, *, timeout: int | float | None = None) -> None:
```
frederikaalund, line 385: Good idea to have method-local overrides to `timeout`. :+1: 

I propose that we use `timeout: float | None = None` (see the previous comment).
Furthermore, I propose the following semantics (written in psedo-code here):

```python
if timeout is None:
    "Use the instance default (`self._timeout`)"
else:
    "Use the `timeout` argument"
```

asyncio_mqtt/client.py
```diff
@@ -404,15 +406,19 @@ async def connect(self, *, timeout: int = 10) -> None:
         # See: https://github.com/eclipse/paho.mqtt.python/blob/v1.5.0/src/paho/mqtt/client.py#L1770
         except (OSError, mqtt.WebsocketConnectionError) as error:
             raise MqttError(str(error)) from None
-        await self._wait_for(self._connected, timeout=timeout)
+        await self._wait_for(
+            self._connected, timeout=self.timeout if timeout is None else timeout
```
frederikaalund, line 410: Again, good job with the "layered" timeout (use argument if given; otherwise, use instance default). Since we use that everywhere, let's move it inside `_wait_for` itself (to remove redundancy).

Inside `_wait_for` we have:

```python
if timeout is None:
    timeout = self._timeout
# Note that asyncio uses `None` to mean "No timeout". We use `math.inf`.
timeout_for_asyncio = None if timeout is math.inf else timeout
try:
    return await asyncio.wait_for(fut, timeout=timeout_for_asyncio, **kwargs)
except ...
```

asyncio_mqtt/client.py
```diff
@@ -404,15 +406,19 @@ async def connect(self, *, timeout: int = 10) -> None:
...
-    async def disconnect(self, *, timeout: int = 10) -> None:
+    async def disconnect(self, *, timeout: int | float | None = None) -> None:
         rc = self._client.disconnect()
         # Early out on error
         if rc != mqtt.MQTT_ERR_SUCCESS:
             raise MqttCodeError(rc, "Could not disconnect")
         # Wait for acknowledgement
-        await self._wait_for(self._disconnected, timeout=timeout)
+        await self._wait_for(
+            self._disconnected, timeout=self.timeout if timeout is None else timeout
```
frederikaalund, line 420: See previous comment about `_wait_for`.

asyncio_mqtt/client.py
```diff
@@ -441,15 +447,17 @@ async def subscribe(
         ] = asyncio.Future()
         with self._pending_call(mid, callback_result, self._pending_subscribes):
             # Wait for callback_result
-            return await self._wait_for(callback_result, timeout=timeout)
+            return await self._wait_for(
+                callback_result, timeout=self.timeout if timeout is None else timeout
```
frederikaalund, line 451: See previous comment about `_wait_for`.

asyncio_mqtt/client.py
```diff
@@ -460,7 +468,10 @@ async def unsubscribe(
         confirmation = asyncio.Event()
         with self._pending_call(mid, confirmation, self._pending_unsubscribes):
             # Wait for confirmation
-            await self._wait_for(confirmation.wait(), timeout=timeout)
+            await self._wait_for(
+                confirmation.wait(),
+                timeout=self.timeout if timeout is None else timeout,
```
frederikaalund, line 473: See previous comment about `_wait_for`.

asyncio_mqtt/client.py
```diff
@@ -487,7 +498,10 @@ async def publish(
         confirmation = asyncio.Event()
         with self._pending_call(info.mid, confirmation, self._pending_publishes):
             # Wait for confirmation
-            await self._wait_for(confirmation.wait(), timeout=timeout)
+            await self._wait_for(
+                confirmation.wait(),
+                timeout=self.timeout if timeout is None else timeout,
```
frederikaalund, line 503: See previous comment about `_wait_for`.

frederikaalund (changes requested):
Hi Scott, thanks for opening this pull request. :+1: Let me have a look. :)

Great idea to add "layered" timeout (use argument if given; otherwise, use instance default). :+1:

Overall, it looks good. :) I only have minor comments (see the review). Let me know if anything is unclear, or, if you have questions.

~Frederik

frederikaalund:
Credit where credit is due: Use of `math.inf` for "No timeout" inspired by the `anyio` API. :)

asyncio_mqtt/client.py
skewty: I intentionally made `timeout` public. Did you intentionally change it to protected?

frederikaalund: I guess that was out of habit. 😄 We can keep it public. 👍 Just make a comment about it like this:

```python
# This property is intentionally public since it's safe to modify at any point during the lifetime
# of this instance. Aka: There are no race conditions.
self.timeout = timeout
```

Otherwise, I'll probably forget that it is "safe" and think it was left public by accident. 😅

frederikaalund:
Hi skewty, thanks for the changes. 👍 It looks good to me except for one bug: The user can't disable the timeout.

To allow the user to disable the timeout (by setting `timeout=math.inf`), I need this code inside `_wait_for`:
```python
if timeout is None:
    timeout = self._timeout
# Note that asyncio uses `None` to mean "No timeout". We use `math.inf`.
timeout_for_asyncio = None if timeout == math.inf else timeout
try:
    return await asyncio.wait_for(fut, timeout=timeout_for_asyncio, **kwargs)
except ...
```

Does it make sense? Let me know if you want to make the change. Otherwise, I can commit this PR as is and push the change myself. 👍

skewty:
> Does it make sense? Let me know if you want to make the change. Otherwise, I can commit this PR as is and push the change myself. 👍

Please go ahead and merge what is here and make the fixes you want.  🎉
This code has been sitting around long enough.

frederikaalund:
> Please go ahead and merge what is here and make the fixes you want. 🎉
> This code has been sitting around long enough.

Done! Thank you for your contribution to asyncio-mqtt. 👍

PR #197, opened 2023-03-15
Update README.md

AndreasHeine:
not much to say about ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ if sys.platform.lower() == "win32" or os.name.lower() == "nt":
         set_event_loop_policy,
         WindowsSelectorEventLoopPolicy
     )
-    # set the eventloop
+    # set the event loop
     set_event_loop_policy(WindowsSelectorEventLoopPolicy())
 # Run your async application as usual
 asyncio.run(main())
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ If you can't wait for the latest version and want to install directly from GitHu
 Since Python 3.8, the default asyncio event loop is the `ProactorEventLoop`. Said loop [doesn't support the `add_reader` method](https://docs.python.org/3/library/asyncio-platforms.html#windows) that is required by asyncio-mqtt. Please switch to an event loop that supports the `add_reader` method such as the built-in `SelectorEventLoop`:
 
 ```python
-# Change to the "Selector" event loop if plattform is windows
+# Change to the "Selector" event loop if platform is Windows
 if sys.platform.lower() == "win32" or os.name.lower() == "nt":
-    # only import if plattform/os is win32/nt, otherwise "WindowsSelectorEventLoopPolicy" is not present
+    # only import if platform/os is win32/nt, otherwise "WindowsSelectorEventLoopPolicy" is not present
     from asyncio import (
         set_event_loop_policy,
         WindowsSelectorEventLoopPolicy
--- a/README.md
+++ b/README.md
@@ -68,10 +68,18 @@ If you can't wait for the latest version and want to install directly from GitHu
 Since Python 3.8, the default asyncio event loop is the `ProactorEventLoop`. Said loop [doesn't support the `add_reader` method](https://docs.python.org/3/library/asyncio-platforms.html#windows) that is required by asyncio-mqtt. Please switch to an event loop that supports the `add_reader` method such as the built-in `SelectorEventLoop`:
 
 ```python
-# Change to the "Selector" event loop
-asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+# Change to the "Selector" event loop if plattform is windows
+if sys.platform.lower() == "win32" or os.name.lower() == "nt":
+    # only import if plattform/os is win32/nt, otherwise "WindowsSelectorEventLoopPolicy" is not present
+    from asyncio import (
+        set_event_loop_policy,
+        WindowsSelectorEventLoopPolicy
+    )
+    # set the eventloop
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,10 +68,18 @@ If you can't wait for the latest version and want to install directly from GitHu
 Since Python 3.8, the default asyncio event loop is the `ProactorEventLoop`. Said loop [doesn't support the `add_reader` method](https://docs.python.org/3/library/asyncio-platforms.html#windows) that is required by asyncio-mqtt. Please switch to an event loop that supports the `add_reader` method such as the built-in `SelectorEventLoop`:
 
 ```python
-# Change to the "Selector" event loop
-asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+# Change to the "Selector" event loop if plattform is windows
+if sys.platform.lower() == "win32" or os.name.lower() == "nt":
+    # only import if plattform/os is win32/nt, otherwise "WindowsSelectorEventLoopPolicy" is not present
```
frederikaalund, line 73: Change "plattform" to "platform"

README.md
```diff
@@ -68,10 +68,18 @@ If you can't wait for the latest version and want to install directly from GitHu
 Since Python 3.8, the default asyncio event loop is the `ProactorEventLoop`. Said loop [doesn't support the `add_reader` method](https://docs.python.org/3/library/asyncio-platforms.html#windows) that is required by asyncio-mqtt. Please switch to an event loop that supports the `add_reader` method such as the built-in `SelectorEventLoop`:
 
 ```python
-# Change to the "Selector" event loop
-asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+# Change to the "Selector" event loop if plattform is windows
+if sys.platform.lower() == "win32" or os.name.lower() == "nt":
```
frederikaalund, line 72: I would argue that one of these is enough but I guess there are some exotic python configurations that requires both checks.

README.md
```diff
@@ -68,10 +68,18 @@ If you can't wait for the latest version and want to install directly from GitHu
...
+    # only import if plattform/os is win32/nt, otherwise "WindowsSelectorEventLoopPolicy" is not present
+    from asyncio import (
+        set_event_loop_policy,
+        WindowsSelectorEventLoopPolicy
+    )
+    # set the eventloop
```
frederikaalund, line 78: Change "eventloop" to "event loop"

README.md
AndreasHeine: sorry for the typos...

not sure whats the most reliable one!?

frederikaalund: Me neither. :) We'll keep both. :+1:

frederikaalund:
Thank you for your (continued) contribution to asyncio-mqtt. :+1:

README.md
frederikaalund: Good question. :)

> There should be one-- and preferably only one --obvious way to do it.

From https://peps.python.org/pep-0020/#the-zen-of-python

Guess that doesn't really apply in this case! I count three ways to do the same thing and none of them is the obvious choice.

JonathanPlasse: The classic

JonathanPlasse: https://stackoverflow.com/questions/4553129/when-to-use-os-name-sys-platform-or-platform-system

frederikaalund: I had a cursory look at the CPython source.

`sys.platform` defined here: https://github.com/python/cpython/blob/afa6092ee4260bacf7bc11905466e4c3f8556cbb/Python/sysmodule.c#L3145
It's set to a compiled-in constant value (from Python.h) when you initialize the `sys` module (calls `_PySys_InitCore`).

`os.name` defined here (for Windows): https://github.com/python/cpython/blob/afa6092ee4260bacf7bc11905466e4c3f8556cbb/Lib/os.py#L73
Lookup into the `sys.builtin_module_names` list. Said list is [from here](https://github.com/python/cpython/blob/afa6092ee4260bacf7bc11905466e4c3f8556cbb/Python/sysmodule.c#L2296). It seems more like "I loaded this module" flag than a "this is the name of the operating system" variable.

`platform.system()` defined here: https://github.com/python/cpython/blob/afa6092ee4260bacf7bc11905466e4c3f8556cbb/Lib/platform.py#L976
Basically calls `platform.uname()` that, in turn, first tries `os.uname()`, and if it fails then `sys.platform`. Does a bunch of expensive runtime checks (might even open a subprocess!) to precisely determine the platform.

---

In summary:

 * Use `sys.platform` if you want to know what system Python was compiled for. It's a low-level string directly from the compilation step. Very fast to "compute".
 * Use `os.name` if you want to know the "backend" for the `os` module. Note that `java` is a valid value! Not really suitable for consistent system detection IMHO.
 * Use `platform.system()` if you want *runtime* system detection (e.g., what's OS is python running on?). It's an "expensive" call though.

I vote for the latter. :+1:

JonathanPlasse: It is expensive but only done once though.
By the latter, you mean `platform.system()`

frederikaalund: > By the latter, do you mean platform.system()?

Yes, that's the one. :) I favor it because it does the right thing (at the cost of performance). My priority list: Correctness (highest prio), maintainability (second highest), performance (lowest).

*Sometimes performance requirements are part of the feature spec. In this case performance leaks into the correctness category*.

AndreasHeine: runtime performance over startup performance !?

frederikaalund: > runtime performance over startup performance !?

Good point, I didn't make a distinction there. Guess I'm undecided. :) In general: If you want performance, Python (in 2023) is not the solution.

Issue #202, opened 2023-03-29
Renaming asyncio-mqtt to aiomqtt


empicano:
@mossblaser has kindly offered to pass the nicer aiomqtt name on to the asyncio-mqtt project, thank you again! 😎

There are some projects depending on aiomqtt, so if necessary, Jonathan wants to keep the option to publish bugfixes. That does not affect asyncio-mqtt due to the lower version number (aiomqtt is on `0.1.3`).

As for migrating to the new name, my idea would be:

1. Publish the package both under the asyncio-mqtt and the aiomqtt name
2. Change all references in the documentation to aiomqtt and log deprecation warning when asyncio-mqtt instead of aiomqtt is imported
3. Stop publishing to asyncio-mqtt on the first major release

@frederikaalund, @JonathanPlasse what do you think?

frederikaalund:
Hey Felix, really great initiative!

I'm currently on vacation until mid next week. I'll have a more detailed look at that time.

~Frederik
________________________________
From: Felix Böhm ***@***.***>
Sent: Wednesday, March 29, 2023 2:08:44 PM
To: sbtinstruments/asyncio-mqtt ***@***.***>
Cc: Frederik Aalund ***@***.***>; Mention ***@***.***>
Subject: [sbtinstruments/asyncio-mqtt] Renaming asyncio-mqtt to aiomqtt (Issue #202)


@mossblaser<https://github.com/mossblaser> has kindly offered to pass the nicer aiomqtt name on to the asyncio-mqtt project, thank you again! 😎

There are some projects depending on aiomqtt, so if necessary, Jonathan wants to keep the option to publish bugfixes. That does not affect asyncio-mqtt due to the lower version number (aiomqtt is on 0.1.3).

As for migrating to the new name, my idea would be:

  1.  Publish the package both under the asyncio-mqtt and the aiomqtt name
  2.  Change all references in the documentation to aiomqtt and log deprecation warning when asyncio-mqtt instead of aiomqtt is imported
  3.  Stop publishing to asyncio-mqtt on the first major release

@frederikaalund<https://github.com/frederikaalund>, @JonathanPlasse<https://github.com/JonathanPlasse> what do you think?

frederikaalund:
Alright, I'm back. :) Sorry for the delay.

Pros:

 * Reduces the number of libraries in the "async mqtt implementations in Python" space. There are plenty already so convergence is a good thing IMHO.

Cons:

[...]

PR #210, opened 2023-05-01
Switch to poetry and rename to aiomqtt

empicano:
Finally found some time for this! @frederikaalund, @JonathanPlasse, if you could look over the `pyproject.toml` that would be great, in particular, I'm not sure if I configured the [`poetry-dynamic-versioning`](https://github.com/mtkennerly/poetry-dynamic-versioning) right.

I published it on Test PyPI: https://test.pypi.org/project/aiomqtt-test/0.16.1.post59

In this PR I still want to update all references from asyncio-mqtt to aiomqtt in the documentation and the code. When the PR is merged I would release `0.17.0` both under the aiomqtt and the asyncio-mqtt names. For asyncio-mqtt I would add a log warning explaining the rename and the migration and publish it manually. Then I would change the name of the repository on GitHub (which also updates the URL of the documentation).

Contrary to what I proposed in #202 I would make this the last release for asyncio-mqtt. The rename doesn't break anything for asyncio-mqtt users that want to stay with the old name, and the migration should be straightforward. Also, I admit that I don't know how to publish to two names automatically and expect this to be unnecessarily complex.

Let me know what you think! 😊


JonathanPlasse:
I am busy this week.
It will take some time for me to review this PR.

.python-version
```diff
@@ -0,0 +1 @@
+3.10
```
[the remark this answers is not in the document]
empicano: I want to pin the Python version and the dependencies for development. This is how I usually do it, but I'm certainly not a pro with this, so if there's a better way let me know :slightly_smiling_face:

pyproject.toml
```diff
@@ -1,70 +1,60 @@
...
-    "furo>=2022.9.29",
-    "sphinx-autobuild>=2021.3.14",
-    "myst-parser>=0.18.1",
-    "sphinx-copybutton>=0.5.1",
-]
-
-[tool.setuptools] # https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
-packages = ["asyncio_mqtt"]
+[tool.poetry.dependencies]
+python = "^3.8"
```
JonathanPlasse, line 23: Are we dropping support of python 3.7?

pyproject.toml
```diff
@@ -1,70 +1,60 @@
...
-    "Operating System :: Microsoft :: Windows",
-    "Operating System :: MacOS",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.7",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
-    "Programming Language :: Python :: Implementation :: CPython",
-    "Programming Language :: Python :: Implementation :: PyPy",
```
JonathanPlasse, line 30: Why remove these classifiers?

.python-version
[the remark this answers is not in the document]
JonathanPlasse: All good, that how I usually do it too.

pyproject.toml
empicano: Good catch, that was not intended 👍

pyproject.toml
empicano: Poetry adds those automatically from the version number! 🙂

JonathanPlasse: Wow, awesome !

.python-version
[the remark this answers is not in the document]
frederikaalund: I usually let poetry decide for itself based on `pyproject.toml`. Right now, it says `3.8` in `pyproject.toml` and `3.10` in `.python-version`. Which one is it? :)

In turn, I check in a `poetry.toml` file with:

```toml
[virtualenvs]
create = true
in-project = true
```

With that, poetry created a local `.venv` folder (instead of in your user directory).

What do you guys think of that approach?

pyproject.toml
```diff
@@ -1,70 +1,60 @@
...
-
-[project]
-name = "asyncio_mqtt"
-description = "Idiomatic asyncio wrapper around paho-mqtt"
+[tool.poetry]
+name = "aiomqtt-test"
+version = "0.0.0"  # Placeholder: Managed by poetry-dynamic-versioning
+description = "The idiomatic asyncio MQTT client, wrapped around paho-mqtt"
+license = "BSD-3-Clause"
+authors = ["Frederik Aalund <fpa@sbtinstruments.com>"]
```
frederikaalund, line 6: Add yourself here as well! :)

Let's just make `authors` the same as `maintainers`. That's how it is in practice.

pyproject.toml
```diff
@@ -1,70 +1,60 @@
...
-    "myst-parser>=0.18.1",
-    "sphinx-copybutton>=0.5.1",
-]
-
-[tool.setuptools] # https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
-packages = ["asyncio_mqtt"]
+[tool.poetry.dependencies]
+python = "^3.8"
+paho-mqtt = ">=1.6.0"
+typing-extensions = {version = ">=4.4.0", markers = "python_version < \"3.10\""}
```
frederikaalund, line 25: Small thing, but let's just use:

```toml
typing-extensions = {version = ">=4.4.0", markers = "python_version < '3.10'"}
```

To avoid the escape characters. :)

pyproject.toml
```diff
@@ -1,70 +1,60 @@
...
+[tool.poetry.group.dev.dependencies]
+mypy = "^1.2.0"
+ruff = "^0.0.263"
+types-paho-mqtt = "^1.6.0.6"
+black = "^23.3.0"
+pytest = "^7.3.1"
+pytest-cov = "^4.0.0"
+anyio = "^3.6.2"
+sphinx = "^6.2.1"
+furo = "^2023.3.27"
```
frederikaalund, line 39: Optional: Make a `tool.poetry.group.docs.dependencies` for these documentation-specific dependencies. :)

pyproject.toml
```diff
@@ -1,70 +1,60 @@
...
+[build-system]
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
+build-backend = "poetry_dynamic_versioning.backend"
 
 [tool.black] # https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file
 target-version = ["py37", "py38", "py39", "py310", "py311"]
 
-[tool.isort] # https://pycqa.github.io/isort/docs/configuration/options.html
-color_output = true
-profile = "black"
```
frederikaalund, line 66: Should we keep `profile = "black"`? :) I have it in my other projects. Maybe it's not in use anymore, or?

frederikaalund (changes requested):
First of all, thank you Felix for taking your time to look into this! :)

Overall it looks great. :+1: I only have some minor comments. Feel free to ask questions. I'll try to elaborate in a timely manner.

I agree with your asyncio-mqtt to aiomqtt rename strategy. :+1: Again, let me know if you need assistance with that. :)

---

Optional: Prefer "standard" [version scheme of PEP 440](https://peps.python.org/pep-0440/#version-matching) for the dependency specifications (e.g., no carets and tildes). PEP 440 is part of PEP 621 and the latter aims to standardize the dependency specification across project management/build systems (poetry, hatch, setuptools, etc.). Might as well hop on board already. :)

pyproject.toml
empicano: I believe [ruff includes the isort functionality](https://beta.ruff.rs/docs/settings/#isort). isort is also no longer in our dev dependencies so I think those options aren't used anymore, and we just forgot to remove them

.python-version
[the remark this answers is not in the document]
empicano: > What do you guys think of that approach?

Wow, I didn't know that existed, that's well hidden in the documentation! Fine by me 👍

> Which one is it?

Both! 😊 The `pyproject.toml` specifies _`^3.8`_, which includes `3.10`. I'd like to have reproducible builds of the documentation (and formatting/linting), that's why I'd like to pin the development dependencies (in `poetry.lock`) and the Python version somehow. This way the GitHub workflow to build the docs can use the exact same ones as we do locally.

I'm not set on this though, do you think there are benefits to not locking the dependencies for development?

empicano:
I'd like to make our `CONTRIBUTING.md` a bit easier by going for GitHub's ["Scripts to rule them all"](https://github.com/github/scripts-to-rule-them-all). We can then also use these scripts inside the GitHub workflows (see `scripts/setup` inside the `docs` workflow). What do you think about that?

empicano:
Does any of you have experience with using `poetry-dynamic-versioning` inside the GitHub workflow and can point me to a code snippet? I'm a bit stuck as to how to do this. I also found this GitHub action that takes the version from the tag on GitHub: https://github.com/marketplace/actions/pypi-poetry-publish

pyproject.toml
frederikaalund: Right! I completely missed that. 😅  I've yet to convert my others projects to use ruff. :)

.python-version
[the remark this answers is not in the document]
frederikaalund: That makes sense. :) It works for me.

frederikaalund:
> I'd like to make our `CONTRIBUTING.md` a bit easier by going for GitHub's ["Scripts to rule them all"](https://github.com/github/scripts-to-rule-them-all). We can then also use these scripts inside the GitHub workflows (see `scripts/setup` inside the `docs` workflow). What do you think about that?

Great idea! 👍 That is one of the more difficult things to maintain manually. PRs usually forget about it or find that its a barrier for contribution.

> Does any of you have experience with using poetry-dynamic-versioning inside the GitHub workflow and can point me to a code snippet?

I'm afraid not. I tried it way back (in another project) but never got a workflow up around it.

frederikaalund (approved):

empicano:
@JonathanPlasse what does this JUnit stuff [here](https://github.com/sbtinstruments/asyncio-mqtt/blob/52a90b6b192df281a8d688a72d653ae680e946ea/pyproject.toml#L117) and in the test workflow do? Do we still need this?

JonathanPlasse:
Yes, we still need this.
It is used to have error messages in the PR code where there are `mypy` or `pytest` errors.

empicano:
pre-commit is really starting to annoy me, I just want to test the workflows without pulling in the changes from main

JonathanPlasse:
You can use [`workflow_dispatch`](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) to manually run the workflows.

frederikaalund:
Great to see progress on this issue. 👍 Keep up the good work! :)

I have some time the next couple of days to contribute as well. Let me know if you need something from my side.

.github/workflows/test.yml
```diff
@@ -22,32 +22,37 @@ jobs:
...
         with:
           name: ${{ matrix.os }} Python ${{ matrix.python-version }}
           files: ./reports/coverage.xml
       - name: Publish JUnit reports
         uses: mikepenz/action-junit-report@v3
-        if: always() # always run even if the previous step fails
+        if: always() # Always run, even if the previous step fails
         with:
           check_name: Junit reports
           report_paths: "./reports/*.xml"
```
JonathanPlasse, line 58: This should be updated as you removed
```toml
[tool.coverage.xml] # https://coverage.readthedocs.io/en/latest/config.html#xml
output = "reports/coverage.xml"
```

empicano: Good catch, that was not on purpose 👍

empicano:
> You can use workflow_dispatch to manually run the workflows.

Maybe I'm not seeing something, but I can only run the workflows from `main` there.

I'm a bit frustrated with changing the tooling to poetry. This takes a lot more time than I anticipated. Can we remove pre-commit in favor of a script (see `scripts/check` that runs ruff and mypy) that we can run locally and that runs on each push and PR? I feel that pre-commit is really slowing me down (e.g. see above, there's no reason this should block the other workflows, but also I want to run the checks when _I_ want, not necessarily on each commit).

Related, the idea to "enable the strict rules and then help contributors fix them" from #217 doesn't work if pre-commit forces contributors to fix things before they commit.

If we remove pre-commit I think I can finish this PR with some oversight from Jonathan on the JUnit stuff. Otherwise, I have to pass this on to someone else because I'm missing the time. Independently, I can still update the docs to the new name.

frederikaalund:
It's fine with me to disable pre-commit to get the move to poetry done. 👍 If I had to choose between the two, then poetry wins no doubt. :) We have all the checks server-side (via GitHub workflows) anyhow.

The idea with pre-commit is that you run the same checks locally first to avoid the wait time (i.e., the wait time for the GitHub workflows to complete). :) That is a nice-to-have feature. We can disable it for now to get things moving.

On a side note: Now that we have a poetry-based local environment, we can use that for pre-commit. That is, use `repo: local` in the `.pre-commit-config.yaml` for all python-based "hooks". See an example of that type of config here: https://github.com/sbtinstruments/cyto/blob/master/.pre-commit-config.yaml
This way, we make `pyproject.toml` file for all our python dependencies (even dev dependencies). Otherwise, pre-commit makes separate envs for it's hooks (and different versions).

frederikaalund:
Also, there is always `git commit --no-verify` if you wan to skip the pre-commit hooks. They are *optional* in that sense. We can't enforce the client to apply said hooks. :)

frederikaalund:
> Anyways, if I interpret it right, you're for keeping pre-commit as a workflow (which is what is blocking me from testing the updated workflows atm), so I'll focus on replacing all occurrences of "asyncio-mqtt" with "aiomqtt" in the docs, and let someone else finish the move to poetry 😋

Ah, I actually didn't realize that we use pre-commit as a driver for our GitHub workflows. 😅

Well, let's get the rename done with. Then we'll tackle the move the poetry in another PR. :) 👍

frederikaalund:
I just looked the GitHub workflow files. I can't see any reference to pre-commit there. Am I missing something?

empicano:
Screenshot 2023-05-19 at 21 23 21

Yes, and who knows why, but pre-commit blocks the execution of other workflows because there are merge commits (even though this is still a draft). I can't look into the project settings, but I suspect that pre-commit is listed as an authorized app there and that this is how it has access.

JonathanPlasse:
This is a GitHub app which must be disabled.

JonathanPlasse:
You should be able to disable it [Here](https://github.com/marketplace/pre-commit-ci)

empicano:
Ping @frederikaalund in case this got lost 😉

frederikaalund:
> Ping @frederikaalund in case this got lost 😉

Sorr about that! 😅 It did get lost among my other work.

I just suspended the pre-commit app. Let me know if you need something more! 👍

empicano:
@JonathanPlasse, any idea why the PyPy tests keep failing? The other tests work fine, and I can't wrap my head around the `Directory /opt/hostedtoolcache/PyPy/3.9.16/x64/lib/pypy3.9 for cffi does not seem to be a Python package` error.

mossblaser:
Hah, glad to hear it led to less hacky things!

I'd welcome a suitable message to be inserted into my old aiomqtt repo's readme pointing people to your project if you'd be happy to suggest some suitable words?

frederikaalund:
> I'd welcome a suitable message to be inserted into my old aiomqtt repo's readme pointing people to your project if you'd be happy to suggest some suitable words?

I already like the current [You might prefer aiomqtt](https://github.com/mossblaser/aiomqtt#you-might-prefer-asyncio-mqtt) section. :) Thank you for that, btw.

.github/workflows/publish.yml
```diff
@@ -7,18 +7,32 @@ on:
...
-      - name: Install dependencies
-        run: pip install build
-      - name: Create package
-        run: python -m build -s -w .
-      - name: Upload package
+          cache: "poetry"
+      - name: Setup dynamic versioning
+        run: poetry self add "poetry-dynamic-versioning[plugin]"
+      - name: Build package
+        run: poetry build
```
webknjaz, line 36: @empicano FYI, building in the same job as publishing is insecure. The build toolchain mustn't have access to OIDC privilege: https://github.com/marketplace/actions/pypi-publish#non-goals.

PR #213, opened 2023-05-15
Fix/some type hints

vvanglro:
Better type hints.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -105,7 +105,7 @@ SubscribeTopic: TypeAlias = (
 )
 
 P = ParamSpec("P")
-S = TypeVar("S", bound="Client")
+S = TypeVar("S", bound="ClientT")
 
 
 # TODO: Simplify the logic that surrounds `self._outgoing_calls_sem` with
@@ -115,7 +115,7 @@ def _outgoing_call(
     method: Callable[Concatenate[S, P], Coroutine[Any, Any, T]]
 ) -> Callable[Concatenate[S, P], Coroutine[Any, Any, T]]:
     @functools.wraps(method)
-    async def decorated(self: S, /, *args: P.args, **kwargs: P.kwargs) -> T:
+    async def decorated(self: S, *args: P.args, **kwargs: P.kwargs) -> T:
         if not self._outgoing_calls_sem:
             return await method(self, *args, **kwargs)
 
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -32,6 +32,10 @@ from typing import (
 if sys.version_info >= (3, 10):
     from typing import Concatenate, ParamSpec, TypeAlias
 else:
+    if sys.version_info >= (3, 11):
+        from typing import Self
+    else:
+        from typing_extensions import Self
     from typing_extensions import Concatenate, ParamSpec, TypeAlias
 
 from contextlib import asynccontextmanager
@@ -860,7 +864,7 @@ class Client:
         while self._client.loop_misc() == mqtt.MQTT_ERR_SUCCESS:
             await asyncio.sleep(1)
 
-    async def __aenter__(self) -> "Client":
+    async def __aenter__(self) -> Self:
         """Connect to the broker."""
         await self.connect()
         return self
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -31,11 +31,11 @@ from typing import (
 
 if sys.version_info >= (3, 10):
     from typing import Concatenate, ParamSpec, TypeAlias
-else:
     if sys.version_info >= (3, 11):
         from typing import Self
     else:
         from typing_extensions import Self
+else:
     from typing_extensions import Concatenate, ParamSpec, TypeAlias
 
 from contextlib import asynccontextmanager
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -547,7 +547,7 @@ class Client:
     async def messages(
         self,
         *,
-        queue_class: Type[asyncio.Queue[Message]] = asyncio.Queue,
+        queue_class: "Type[asyncio.Queue[Message]]" = asyncio.Queue,
         queue_maxsize: int = 0,
     ) -> AsyncGenerator[AsyncGenerator[Message, None], None]:
         """Return async generator of incoming messages.
@@ -622,7 +622,7 @@ class Client:
     def _callback_and_generator(
         self,
         *,
-        queue_class: Type[asyncio.Queue[Message]] = asyncio.Queue,
+        queue_class: "Type[asyncio.Queue[Message]]" = asyncio.Queue,
         queue_maxsize: int = 0,
     ) -> Tuple[Callable[[Message], None], AsyncGenerator[Message, None]]:
         # Queue to hold the incoming messages
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: BSD-3-Clause
-from __future__ import annotations
 
 import asyncio
 import functools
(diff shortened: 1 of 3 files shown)


JonathanPlasse:
What is the motivation for those changes?

frederikaalund:
> What is the motivation for those changes?

Seems to be for python 3.7 support. :) I personally welcome this effort! :+1: At least until we officially drop support for 3.7. What do you guys (asyncio-mqtt authors/maintainers) think? It *is* a step backwards so an alternative to this PR is to drop support for old versions of python.

@vvanglo thanks for another great PR! We (the authors/maintainers) have to align on the direction first.

vvanglro:
I think this is a general method of type hinting, whether it is python3.7 or later.  And the delay detection is removed. Personally, I don't like to use delay detection very much.

vvanglro:
There is another motivation because of this. 😄 
Tips for pylance in vscode:
image
image

empicano:
> What do you guys (asyncio-mqtt authors/maintainers) think?

I'm not a good authority on typing because I don't use it that much, I'm fine with both versions 👍

Apart from that, [Python 3.7 is almost at end-of-life](https://devguide.python.org/versions/) so I think it's reasonable to drop support as well if it's extra effort for us.

JonathanPlasse:
> There is another motivation because of this. smile Tips for pylance in vscode: image image

It should just not be split into multiple lines.

vvanglro:
> > There is another motivation because of this. smile Tips for pylance in vscode: image image
> 
> It should just not be split into multiple lines.

Beautiful misunderstanding😂

frederikaalund:
> I think this is a general method of type hinting, whether it is python3.7 or later

I wish that was true! However, the "recommended" approach to type annotations change depending on the target python version. At least in my opinion. Here is an quick primer:

#### Untyped (python 3.4 and below)
```python
self._pending_subscribes = {}
```

#### Typed (python 3.5; when we first got the `typing` module)
```python
import paho.mqtt.client as mqtt
import asyncio
from typing import Dict, List, Tuple  # These were deprecated in 3.9 in favor of `dict`, `list` and `tuple`.
from typing import Union

self._pending_subscribes: Dict[int, asyncio.Future[Union[Tuple[int, ...], List[mqtt.ReasonCodes]]]] = {}
```
Quite verbose! 😅 

#### Typed (python 3.9; built-ins now supports subscripting via `[]`)
```python
import paho.mqtt.client as mqtt
import asyncio
from typing import Union

self._pending_subscribes: dict[int, asyncio.Future[Union[tuple[int, ...], list[mqtt.ReasonCodes]]]] = {}
```
Looks a bit more natural but still quite verbose.

#### Typed (python 3.10; Unions via `|`)
```python
import paho.mqtt.client as mqtt
import asyncio
from typing import Union

self._pending_subscribes: dict[int, asyncio.Future[tuple[int, ...] | list[mqtt.ReasonCodes]]] = {}
```
This is the state of the art. :) At least as far as I know. 😅 

Sooo, if we want to do state-of-the-art typing, we need Python 3.10 and above. Dropping support for python 3.7 alone doesn't really give us much in term of typing.

We can do tricks to get new-style type annotations in older python versions. E.g., to wrap the annotations in strings.

Basically, what I see that **this PR accomplishes is to make asyncio-mqtt compatible with python 3.7** via the use of old-style (e.g., pre-3.9) typing. I say 3.7 (and not 3.5) because we use other, non-typing features of 3.7 so that's the minimum version as of now.

Am I missing something or is this the intend of this PR? :) Let me know if I'm wrong!

---

> And the delay detection is removed. Personally, I don't like to use delay detection very much.

What do you mean by this mean? :) I can't find the relevant code change (it's all typing related).

vvanglro:
Yes, I went and read PEP585 and currently the project is using the latest type hinting methods from py39 upwards.

List and Dict in typing will be removed until the end of py39 life. At least this is the plan in the PEP585 proposal. 

At present, there are still types such as List Dict Union in py37-11, So I say this is a general type hint.

I call it latency detection:
```py
from __future__ import annotations
```

frederikaalund:
> I call it latency detection:
> 
> ```python
> from __future__ import annotations
> ```

Ah, now it makes sense. I guess that we used it as a workaround. That is, so we can have nice, state-of-the-art typing (ala 3.9 and above) while being backwards-compatible with 3.7. It's not consistent though. I wonder if `from __future__ import annotations` has the same effect (gives us backwards compatibility but without the strings). That would be a lot cleaner!

vvanglro:
`from __future__ import annotations`

Importing it only works in this case:

image

Import it and include type hints with strings to make some higher version hints available for lower versions.

But this is only for static type checking, Doesn't work at runtime. 

It is explained in this article: https://dev.to/tiangolo/the-future-of-fastapi-and-pydantic-is-bright-3pbm

image

Refer to:
1. https://peps.python.org/pep-0563/
2. https://peps.python.org/pep-0649/

frederikaalund:
Thanks for linking the tiangolo article. I remember when I first saw that issue on the pydantic bug tracker way back.

> But this is only for static type checking, Doesn't work at runtime.

Doesn't work *if you do advanced stuff like declaring the class inside a function*. For our use case of "runtime types" via `dataclass` it works just fine. That's because we declare our `dataclass`es on the top level (globally).

In any case, **I vote that we (asyncio-mqtt team)**:
 * Use starte-of-the-art type annotations as per the python 3.10 style (this is the status quo).
 * Keep `from __future__ import annotations` to be backwards compatible with earlier python versions (below 3.10)
 * On a case-by-case basis: Wrap our type annotations in strings to make IDEs, type checkers, linters, etc. happy.

@JonathanPlasse and @empicano: What do you think?

### On the issue of runtime tying

There is no runtime problem for us (asyncio-mqtt) because we have a simple use case for runtime typing: top-level `dataclass`es. E.g.:

```python
@dataclass(frozen=True)
class TLSParameters:
    ca_certs: str | None = None  # python 3.10 style type annotations
    ...
```

I just did a test on python 3.7 to verify that the type hints are indeed backwards compatible:
```shell
fpa@sbt-fpa-desktop1:/projects/asyncio-mqtt$ poetry run python -c "import sys; print(sys.version); from asyncio_mqtt.client import TLSParameters; x = TLSParameters(ca_certs='something'); print(x)"
3.7.16 (default, Dec  7 2022, 01:12:19) 
[GCC 9.4.0]
TLSParameters(ca_certs='something', certfile=None, keyfile=None, cert_reqs=None, tls_version=None, ciphers=None, keyfile_password=None)
```
It works! Python 3.7 magically understands the 3.10-style type hints. Honestly, I don't understand why (I can't find any mention in the python changelog for 3.7x) but I'm glad it works.
If I remove the `from __future__ import annotations` like, I get the expected error:
```
fpa@sbt-fpa-desktop1:/projects/asyncio-mqtt$ poetry run python -c "import sys; print(sys.version); from asyncio_mqtt.client import TLSParameters; x = TLSParameters(ca_certs='something'); print(x)"
3.7.16 (default, Dec  7 2022, 01:12:19) 
[GCC 9.4.0]
  File "<string>", line 1, in <module>
  File "/projects/asyncio-mqtt/asyncio_mqtt/__init__.py", line 3, in <module>
    from .client import (
  File "/projects/asyncio-mqtt/asyncio_mqtt/client.py", line 58, in <module>
    @dataclass(frozen=True)
  File "/projects/asyncio-mqtt/asyncio_mqtt/client.py", line 60, in TLSParameters
    ca_certs: str | None = None
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

JonathanPlasse:
The CI runs for all current Python versions.
It works on Python 3.7 because the types are never evaluated as types but as strings.

frederikaalund:
> It works on Python 3.7 because the types are never evaluated as types but as strings.

Still doesn't explain why the 3.10 style type hints work for the *runtime* `dataclass`es in 3.7. I mean, the `dataclass` must inspect the type (stringified or not) to deduce how to parse it's arguments. It does work in 3.7 but it's an undocumented feature as far as I can see. :)

JonathanPlasse:
All annotations with `from __future__ import annotations` become strings.
Dataclasses do not generally inspect types https://docs.python.org/3/library/dataclasses.html#class-variables.

frederikaalund:
Ah, you're right. 👍 I thought `dataclass`es where more `pydantic.BaseModel`-like than they actually are. 😅 Now it all makes sense. :)

Issue #48, opened 2021-03-01
Publish should not wait for confirmation


0x0a11c0de:
The `publish()` method forces the user to wait for the message to be published. It is perfectly reasonable to publish a message with QOS > 0 when the client is disconnected. For example, on unreliable links, a loop in a thread/task is used to reconnect when the connection goes down. The publisher can continue on publishing at QOS > 0 without knowing anything about the status of the connection. This is not possible with the current `publish()` implementation and puts a bigger implementation burden on the user of this library.

frederikaalund:
Thanks for opening this issue. Let me have a look. :)

> It is perfectly reasonable to publish a message with QOS > 0 when the client is disconnected. For example, on unreliable links, a loop in a thread/task is used to reconnect when the connection goes down. The publisher can continue on publishing at QOS > 0 without knowing anything about the status of the connection.

I never considered that use case. Thanks for bringing it to my attention. :+1: Your question in #46 makes a lot more sense to me now.

To support this, we need to make some API-breaking changes. I'm up for it, though.

### Make `Client` work in "offline" mode

Old behaviour:
 * `Client.publish` only works while online (connected to the server)
 * `Client` can only connect once and only disconnect once. You have to create a new `Client` instance if you want to, e.g., reconnect to the server.
 
New behaviour:
 * `Client.publish` works while offline (not connected to the server)
 * `Client` can connect to/disconnect from the server multiple times. We use the existing `paho.client.reconnect` logic to support this.

An example of the new API that these changes bring with them (based on the "Advanced example"):

```python
async def ensure_connection(client):
    reconnect_interval = 3  # [seconds]
    while True:
        try:
            # Connect to the server
            async with client:
                # We pass the (connected) client to the advanced example
                await advanced_example(client)
        except MqttError as error:
            print(f'Error "{error}". Reconnecting in {reconnect_interval} seconds.')
        finally:
            await asyncio.sleep(reconnect_interval)

async def post_to_topics(client, topics):
    while True:
        for topic in topics:
            message = randrange(100)
            print(f'[topic="{topic}"] Publishing message={message}')
            # Note that we no longer use await here! If the client isn't connected, paho-mqtt queues the messages
            # up internally.
            client.publish(topic, message, qos=1)
            await asyncio.sleep(2)

# There is only a single client instance. We reuse it again and again.
client = Client("test.mosquitto.org")

# Keep the client connected indefinitely
task = asyncio.create_task(ensure_connection(client))
[...]

Issue #26, opened 2020-11-27
Paho mqtt client reconnecting from callback loop


MartinHjelmare:
## TLDR
There are two cases during connection that will result in reconnecting from the event loop, blocking the loop in the process.

## Background
This library, asyncio-mqtt, wraps paho-mqtt and uses its external loop feature to drive paho-mqtt using the asyncio event loop. The asyncio event loop should never block so it's important to not run blocking I/O inside the event loop.

### paho-mqtt-client details
We hook into the event loop by using the `loop.add_reader` method and connect the paho-mqtt callback `loop_read`. This callback handles the incoming packets from the socket, by calling `_packet_read` and in turn `_packet_handle`. The latter directs to different callbacks depending on the command. One of those cases is the `_handle_connack` callback. There are two cases in this callback that will result in calling the `reconnect` method of the paho mqtt client. This method does blocking I/O. So if we would end up in either of those cases, we would block the event loop during the reconnection.

Case 1:
```py
protocol == MQTTv311 and result == CONNACK_REFUSED_PROTOCOL_VERSION
```

Case 2:
```py
protocol == MQTTv311 and result == CONNACK_REFUSED_IDENTIFIER_REJECTED and self._client_id == b''
```

https://github.com/eclipse/paho.mqtt.python/blob/42f0b13001cb39aee97c2b60a3b4807314dfcb4d/src/paho/mqtt/client.py#L2985-L3003

## Example log with forced incorrect MQTT protocol (case 1)

```
2020-11-27 12:04:31,471 INFO (MainThread) [asyncio_mqtt] Starting client example.
2020-11-27 12:04:31,472 DEBUG (MainThread) [asyncio] Using selector: EpollSelector
2020-11-27 12:04:31,480 DEBUG (ThreadPoolExecutor-0_0) [asyncio_mqtt] Sending CONNECT (u0, p0, wr0, wq0, wf0, c1, k60) client_id=b''
2020-11-27 12:04:31,482 DEBUG (MainThread) [asyncio_mqtt] Received CONNACK (0, 1), attempting downgrade to MQTT v3.1.
2020-11-27 12:04:31,482 DEBUG (MainThread) [asyncio_mqtt] Sending CONNECT (u0, p0, wr0, wq0, wf0, c1, k60) client_id=b''
Error "Operation timed out". Reconnecting in 3 seconds.
[...]

PR #216, opened 2023-05-16
fix: issue212

### LINKED ISSUE #212: Can't recover after connection lost
I can't understand why I can't recover after a connection lost.

```python

import asyncio_mqtt as aiomqtt


client = aiomqtt.Client(
    hostname=broker_address,  # The only non-optional parameter
    port=broker_port
)

# Connect, reconnect, subscribe and check messages
async def mqttConnect():
    interval = 5  # Seconds
    # Run for ever
    while True:
        try:
            async with client as _client:
                async with _client.messages(queue_class=PriorityQueue) as messages:
                    await _client.subscribe("driver/+/speed/set/w")
                    await _client.subscribe("driver/+/cmd/power")
                    await _client.subscribe("driver/+/cmd/control")  
                        await handle_message(message)

        except aiomqtt.MqttError as e:
            print(f'Connection lost; Reconnecting in {interval} seconds ... {e}')
            await asyncio.sleep(interval)

async def main():
    # Create MQTT connection
    # await client.connect()
    # Create MQTT
    asyncio.create_task(mqttConnect(), name="MQTT")
    # Stay here for ever
    while True:
        await asyncio.sleep(1000)

# start the event loop
if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    try:
        loop.run_until_complete(main())
    except KeyboardInterrupt:
        print("Program exiting...")
```

After restarting my mosquitto server I can't recover:

```
Connection lost; Reconnecting in 5 seconds ... [code:7] The connection was lost.
Connection lost; Reconnecting in 5 seconds ... [code:7] The connection was lost.
Connection lost; Reconnecting in 5 seconds ... [code:7] The connection was lost.
```

I've changed the code trying to trace it:

```python
async def mqttConnect():
    interval = 5  # Seconds
    # Run for ever
    while True:
        try:
            print(f"Trying to connect...")
            await client.connect()
            async with client.messages(queue_class=PriorityQueue) as messages:
                await client.subscribe("driver/+/speed/set/w")
                await client.subscribe("driver/+/cmd/power")
                await client.subscribe("driver/+/cmd/control")  
                    await handle_message(message)

        except aiomqtt.MqttError as e:
            print(f'Connection lost; Reconnecting in {interval} seconds ... {e}')
            await asyncio.sleep(interval)
```

```
Connection lost; Reconnecting in 5 seconds ... Disconnected during message iteration
Trying to connect...
Connection lost; Reconnecting in 5 seconds ... Disconnected during message iteration
Trying to connect...
Connection lost; Reconnecting in 5 seconds ... Disconnected during message iteration
Trying to connect...
Connection lost; Reconnecting in 5 seconds ... Disconnected during message iteration
```

Any idea?

vvanglro:
This PR fixes #212 .
If `_disconnected` is found to be completed after connecting, it is not correct, so we need to reset it. 
If `_connected` is still in the completed state after disconnection, reset it.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -34,7 +34,7 @@ from contextlib import asynccontextmanager
 import paho.mqtt.client as mqtt
 from paho.mqtt.properties import Properties
 
-from .error import MqttCodeError, MqttConnectError, MqttError
+from .error import MqttCodeError, MqttConnectError, MqttError, MqttReentrantError
 from .types import PayloadType, T
 
 MQTT_LOGGER = logging.getLogger("mqtt")
@@ -273,6 +273,10 @@ class Client:
         websocket_headers: WebSocketHeaders | None = None,
         max_inflight_messages: int | None = None,
         max_queued_messages: int | None = None,
+        message_queue_class: type[
+            asyncio.Queue[Message | mqtt.MQTTMessage]
+        ] = asyncio.Queue,
+        message_queue_maxsize: int = 0,
     ) -> None:
         self._hostname = hostname
         self._port = port
@@ -373,6 +377,10 @@ class Client:
         if socket_options is None:
             socket_options = ()
         self._socket_options = tuple(socket_options)
+        self._lock: asyncio.Lock = asyncio.Lock()
+        self._message_queue_class: asyncio.Queue[
+            Message | mqtt.MQTTMessage
+        ] = message_queue_class(message_queue_maxsize)
 
     @property
     def id(  # noqa: A003 # TODO: When doing BREAKING CHANGES rename to avoid shadowing builtin id
@@ -423,6 +431,16 @@ class Client:
             self._disconnected: asyncio.Future[int | mqtt.ReasonCodes | None] = asyncio.Future()  # type: ignore[no-redef]
 
     async def disconnect(self, *, timeout: int = 10) -> None:
+        """Disconnect from the broker."""
+        # Early out if already disconnected...
+        if self._disconnected.done():
+            disc_exc = self._disconnected.exception()
+            if disc_exc is not None:
+                # ...by raising the error that caused the disconnect
+                raise disc_exc
+            # ...by returning since the disconnect was intentional
+            return
+        # Try to gracefully disconnect from the broker
         rc = self._client.disconnect()
         # Early out on error
         if rc != mqtt.MQTT_ERR_SUCCESS:
@@ -511,7 +529,7 @@ class Client:
 
     @asynccontextmanager
     async def filtered_messages(
-        self, topic_filter: str, *, queue_maxsize: int = 0
+        self, topic_filter: str
     ) -> AsyncGenerator[AsyncGenerator[mqtt.MQTTMessage, None], None]:
         """Return async generator of messages that match the given filter."""
         self._logger.warning(
@@ -519,7 +537,7 @@ class Client:
             " Use messages() together with Topic.matches() instead."
         )
         callback, generator = self._deprecated_callback_and_generator(
-            log_context=f'topic_filter="{topic_filter}"', queue_maxsize=queue_maxsize
+            log_context=f'topic_filter="{topic_filter}"'
         )
         try:
             self._client.message_callback_add(topic_filter, callback)
@@ -543,7 +561,7 @@ class Client:
             msg = "Only a single unfiltered_messages generator can be used at a time."
             raise RuntimeError(msg)
         callback, generator = self._deprecated_callback_and_generator(
-            log_context="unfiltered", queue_maxsize=queue_maxsize
+            log_context="unfiltered"
         )
         try:
             self._unfiltered_messages_callback = callback
@@ -556,9 +574,6 @@ class Client:
     @asynccontextmanager
     async def messages(
         self,
-        *,
-        queue_class: type[asyncio.Queue[Message]] = asyncio.Queue,
-        queue_maxsize: int = 0,
     ) -> AsyncGenerator[AsyncGenerator[Message, None], None]:
         """Return async generator of incoming messages.
 
@@ -566,9 +581,7 @@ class Client:
         incoming messages will be discarded (and a warning is logged).
         If queue_maxsize is less than or equal to zero, the queue size is infinite.
         """
-        callback, generator = self._callback_and_generator(
-            queue_class=queue_class, queue_maxsize=queue_maxsize
-        )
+        callback, generator = self._callback_and_generator()
         try:
             # Add to the list of callbacks to call when a message is received
             self._on_message_callbacks.append(callback)
@@ -579,20 +592,17 @@ class Client:
             self._on_message_callbacks.remove(callback)
 
     def _deprecated_callback_and_generator(
-        self, *, log_context: str, queue_maxsize: int = 0
+        self, *, log_context: str
     ) -> tuple[
         Callable[[mqtt.Client, Any, mqtt.MQTTMessage], None],
         AsyncGenerator[mqtt.MQTTMessage, None],
     ]:
-        # Queue to hold the incoming messages
-        messages: asyncio.Queue[mqtt.MQTTMessage] = asyncio.Queue(maxsize=queue_maxsize)
-
         # Callback for the underlying API
         def _put_in_queue(
             client: mqtt.Client, userdata: Any, message: mqtt.MQTTMessage
         ) -> None:
             try:
-                messages.put_nowait(message)
+                self._message_queue_class.put_nowait(message)
             except asyncio.QueueFull:
                 self._logger.warning(
                     f"[{log_context}] Message queue is full. Discarding message."
@@ -605,9 +615,7 @@ class Client:
                 # Wait until we either:
                 #  1. Receive a message
                 #  2. Disconnect from the broker
-                get: asyncio.Task[mqtt.MQTTMessage] = self._loop.create_task(
-                    messages.get()
-                )
+                get: asyncio.Task[mqtt.MQTTMessage] = self._loop.create_task(self._message_queue_class.get())  # type: ignore[arg-type]
                 try:
                     done, _ = await asyncio.wait(
                         (get, self._disconnected), return_when=asyncio.FIRST_COMPLETED
@@ -631,17 +639,12 @@ class Client:
 
     def _callback_and_generator(
         self,
-        *,
-        queue_class: type[asyncio.Queue[Message]] = asyncio.Queue,
-        queue_maxsize: int = 0,
     ) -> tuple[Callable[[Message], None], AsyncGenerator[Message, None]]:
         # Queue to hold the incoming messages
-        messages: asyncio.Queue[Message] = queue_class(maxsize=queue_maxsize)
-
         def _callback(message: Message) -> None:
             """Put the new message in the queue."""
             try:
-                messages.put_nowait(message)
+                self._message_queue_class.put_nowait(message)
             except asyncio.QueueFull:
                 self._logger.warning("Message queue is full. Discarding message.")
 
@@ -651,7 +654,7 @@ class Client:
                 # Wait until we either:
                 #  1. Receive a message
                 #  2. Disconnect from the broker
-                get: asyncio.Task[Message] = self._loop.create_task(messages.get())
+                get: asyncio.Task[Message] = self._loop.create_task(self._message_queue_class.get())  # type: ignore[arg-type]
                 try:
                     done, _ = await asyncio.wait(
                         (get, self._disconnected), return_when=asyncio.FIRST_COMPLETED
@@ -870,6 +873,10 @@ class Client:
 
     async def __aenter__(self) -> Client:
         """Connect to the broker."""
+        if self._lock.locked():
+            msg = "Does not support reentrant"
+            raise MqttReentrantError(msg)
+        await self._lock.acquire()
         await self.connect()
         return self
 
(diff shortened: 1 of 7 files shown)


frederikaalund:
Thanks for opening this PR. 👍 Let me have a look. :)

I assume the overall purpose is to make `asyncio_mqtt.Client` [*reusable*](https://docs.python.org/3/library/contextlib.html#reusable-context-managers). I quote the linked docs to save you a click:

> Distinct from both single use and reentrant context managers are “reusable” context managers (or, to be completely explicit, “reusable, but not reentrant” context managers, since reentrant context managers are also reusable). These context managers support being used multiple times, but will fail (or otherwise not work correctly) if the specific context manager instance has already been used in a containing with statement.

**That's the purpose of this PR, right? Let me know if I'm wrong.**

If I'm right, then this PR also relates to #48. I'll provide more feedback after we settle on the intend of this PR.

I hope that makes sense. :)

vvanglro:
Yes, after adding the judgment, it will become a reusable context. I read #48, and I don’t know much about the publish and QOS in the question. What I can only confirm is that this PR will make aiomqtt a reusable of the context.

vvanglro:
To be honest, I don't really like to use context in projects, such as the example used in fastapi in your documentation.
I prefer this: (Of course this is just my personal preference😅)
```py
client = aiomqtt.Client("test.mosquitto.org")

@contextlib.asynccontextmanager
async def lifespan(app):
    await clinet.connect()
    yield
    await client.disconnect()
```
I found that you mentioned paho.client.reconnect in #48. 
It would be great if we could integrate this method. See this method literally means reconnect. My good wish is that we can use this method to do More things, such as when the connection is disconnected when calling methods such as subscribing and getting messages, then we will automatically try to reconnect for the user, and the number of reconnections is configurable.
This can be very useful in long-running projects, such as web projects.

frederikaalund:
> Yes, after adding the judgment, it will become a reusable context. I read https://github.com/sbtinstruments/asyncio-mqtt/issues/48, and I don’t know much about the publish and QOS in the question. What I can only confirm is that this PR will make aiomqtt a reusable of the context.

Great, because that's what we need to test for then! :) What I need from you to merge this PR:
 * Add tests that verify that `asyncio_mqtt.Client` is now reusable (but not reentrant!). Things to consider for the tests:
   * What happens if task A reuses the client while task B listens for messages? Will task B get an exception? Will task B get stuck?
   * What happens if we try to reenter the client? Does it raise an exception? Does it deadlock? Does it work?
   * In general, it's good to both tests for what is supposed to *work* and what is definitely supposed to *not work* (e.g., do we raise proper exceptions that the user can understand?).
 * Change/add to the docs that you can now reuse the client. People will like this feature. 👍

The test part is the most important. I can do the docs if you don't feel like it. :)

---

On the topic of "automatic reconnect": Yes! That would be great to have. It is also quite difficult to get right. 😅 See #6 and #26 for some of the subtleties.
This PR (#216) is the first step towards "automatic reconnect". 👍 We can build the latter on top of the former.

vvanglro:
OK, I'll do it.

vvanglro:
This PR changed 5 points:

1. reusable and non reentrant context.
2. Extract the disconnect method.
3. Change message queue to class attribute.
4. Add test case.
5. update ruff version.

asyncio_mqtt/client.py
```diff
@@ -371,6 +377,10 @@ def __init__(  # noqa: C901, PLR0912, PLR0915
         if socket_options is None:
             socket_options = ()
         self._socket_options = tuple(socket_options)
+        self._lock: asyncio.Lock = asyncio.Lock()
```
frederikaalund, line 380: Good idea to use a `Lock`. 👍 It's a great use case for it.

asyncio_mqtt/client.py
```diff
@@ -416,14 +426,30 @@ async def connect(self, *, timeout: int = 10) -> None:
         except (OSError, mqtt.WebsocketConnectionError) as error:
             raise MqttError(str(error)) from None
         await self._wait_for(self._connected, timeout=timeout)
+        # If _disconnected is already completed after connecting, reset it.
+        if self._disconnected.done():
+            self._disconnected: asyncio.Future[int | mqtt.ReasonCodes | None] = asyncio.Future()  # type: ignore[no-redef]
```
frederikaalund, line 431: Skip the type annotation here. We only need it in `__init__` during the member variable declaration. 👍

asyncio_mqtt/client.py
```diff
@@ -416,14 +426,30 @@ async def connect(self, *, timeout: int = 10) -> None:
...
+        # Try to gracefully disconnect from the broker
         rc = self._client.disconnect()
         # Early out on error
         if rc != mqtt.MQTT_ERR_SUCCESS:
             raise MqttCodeError(rc, "Could not disconnect")
         # Wait for acknowledgement
         await self._wait_for(self._disconnected, timeout=timeout)
+        # If _connected is still in the completed state after disconnection, reset it
+        if self._connected.done():
+            self._connected: asyncio.Future[int | mqtt.ReasonCodes | None] = asyncio.Future()  # type: ignore[no-redef]
```
frederikaalund, line 452: Again, skip the type annotation here. :)

asyncio_mqtt/client.py
```diff
@@ -502,15 +528,15 @@ async def publish(  # noqa: PLR0913
 
     @asynccontextmanager
     async def filtered_messages(
-        self, topic_filter: str, *, queue_maxsize: int = 0
+        self, topic_filter: str
```
frederikaalund, line 531: Again, unrelated to this PR. Move into a separate PR. Also, that is a breaking change!

asyncio_mqtt/client.py
```diff
@@ -889,6 +895,8 @@ async def __aexit__(
                 " disconnection."
             )
             await self.force_disconnect()
+        finally:
+            self._lock.release()
```
frederikaalund, line 899: Good use of `finally` 👍

asyncio_mqtt/client.py
```diff
@@ -870,15 +885,6 @@ async def __aexit__(
         exc: BaseException | None,
         tb: TracebackType | None,
     ) -> None:
-        """Disconnect from the broker."""
```
frederikaalund, line 873: Good idea to move this logic into `disconnect`. 👍

asyncio_mqtt/client.py
frederikaalund: Hmm, there are some subtle race conditions in this __aexit__ method now. E.g., if `disconnect` raises, we never call `force_disconnect`. That is bad!

My suggestion:
 * **Breaking change**: Make `force_disconnect` private (call it `_force_disconnect`).
 * Make `_force_disconnect` sync (there is no reason for it to be async anyhow).
 * Move `_force_disconnect` into the `finally` block (just above the lock release)

That should fix the race conditions. Note that `_force_disconnect`: (1) never raises and (2) is a no-op if already disconnected.

Does it make sense?

frederikaalund: Mention the breaking change in the changelog as well. In fact, make a changelog entry about this new reentrant Client feature! 😄 Our users would want to know about it!

tests/test_client.py
```diff
@@ -383,3 +384,123 @@ async def test_client_no_pending_calls_warnings_with_max_concurrent_outgoing_cal
                 tg.start_soon(client.publish, topic)
 
         assert caplog.record_tuples == []
+
+
+async def test_client_not_reentrant() -> None:
+    client = Client(HOSTNAME)
+
+    with pytest.raises(MqttReentrantError):  # noqa: PT012
```
frederikaalund, line 392: Excellent! Great basic test. 👍

tests/test_client.py
```diff
@@ -383,3 +384,123 @@ async def test_client_no_pending_calls_warnings_with_max_concurrent_outgoing_cal
...
+        async with client:
+            async with client:
+                ...
+
+
+async def test_client_reusable() -> None:
```
frederikaalund, line 398: Great. 👍

tests/test_client.py
```diff
@@ -383,3 +384,123 @@ async def test_client_no_pending_calls_warnings_with_max_concurrent_outgoing_cal
...
+        async with custom_client:
+            ...
+
+    async def task_a_publisher() -> None:
+        async with publish_client:
+            await publish_client.publish("task/a", "task_a")
+
+    async with anyio.create_task_group() as tg:
+        tg.start_soon(task_a_customer)
+        await anyio.sleep(1)
```
frederikaalund, line 438: It's an anti-pattern to `sleep` like that to ensure that `task_a_customer` actually started before we proceed. :)

Instead, use `tg.start`. See this great example from the anyio docs: https://anyio.readthedocs.io/en/3.x/tasks.html#starting-and-initializing-tasks

Let me know if you need help with it. 👍

tests/test_client.py
```diff
@@ -383,3 +384,123 @@ async def test_client_no_pending_calls_warnings_with_max_concurrent_outgoing_cal
...
+
+async def test_client_connect_disconnect() -> None:
+    client = Client(HOSTNAME)
+
+    await client.connect()
+    await client.publish("connect", "connect")
+    await client.disconnect()
+
+
+async def test_client_reusable_message() -> None:
```
frederikaalund, line 416: Excellent test! Good scenarios. 👍

frederikaalund (changes requested):
Looks great and thanks for the detailed tests! 👍 :) The more coverage we get on that front the better.

Have a look at the comments in the review. Let me know if anything is unclear.

### Side note for future reference (not directly relevant to this PR)

Looking through the internals like this again makes me wish that I never exposed `connect` and `disconnect` as part of the public interface. 😅 It's quite the maintenance burden!

Also, now we have this mess of internal state (I'm the cause of that myself):

 * `self._connected: asyncio.Future`
 * `self._disconnected: asyncio.Future`
 * `self._lock: asyncio.Lock`

All three variables have semantic overlap! Really, we should just have the single `_lock` (where "lock acquired" means "client connecting/connected"). Extending on that: Have all internal state inside a tagged union (sum type) in rust-style, concurrency-safe semantics.

In any case, all of that is just food for thought for, e.g., an anyio-based rewrite. :)

empicano:
Hi vvanglro, this is a very cool addition, thank you! 🎉 

> [...] makes me wish that I never exposed connect and disconnect as part of the public interface. 😅 It's quite the maintenance burden!

Our documentation didn't document `connect` and `disconnect` for a long time, in fact, it states: "asyncio-mqtt doesn’t support manual calls to `connect` and `disconnect`."

It'd be a breaking change nonetheless, but I think it's valid to remove it as well. I'm always for simplifying things 😄

frederikaalund:
Thank you again for your great work on this PR, @vvanglro. :+1: I plan to do the code review this weekend. I'm really tied up until then. :)

vvanglro:
I'm thinking about some problems,🤯 if we remove connect/disconnect, based main code and the current PR, there will be the following three situations.



The first situation(Base main code, This will establish 10 connections simultaneously):
```py
from asyncio_mqtt import Client


async def run1(message):
    async with Client("192.168.5.24", username="admin", password="123456") as client:
        await client.publish("humidity/outside", payload=message)
        # Maybe it can only be concurrently in context.
        # tasks = [asyncio.create_task(client.publish("humidity/outside", payload=i)) for i in range(10)]
        # await asyncio.gather(*tasks)


async def gather_task1():
    tasks = [asyncio.create_task(run1(i)) for i in range(10)]
    await asyncio.gather(*tasks)
    
    
if __name__ == '__main__':
    import asyncio
    asyncio.get_event_loop().run_until_complete(gather_task1())
```


The two situation(Base current PR, Not reentrant will result in an error):
```py
from asyncio_mqtt import Client


client = Client("192.168.5.24", username="admin", password="123456")


async def run2(message):
    async with client:
        await client.publish("humidity/outside", payload=message)
        # Maybe it can only be concurrently in context.
        # tasks = [asyncio.create_task(client.publish("humidity/outside", payload=i)) for i in range(10)]
        # await asyncio.gather(*tasks)


async def gather_task2():
    tasks = [asyncio.create_task(run2(i)) for i in range(10)]
    await asyncio.gather(*tasks)


if __name__ == '__main__':
    import asyncio
    asyncio.get_event_loop().run_until_complete(gather_task2())
```

The three situation(There is only 1 connection, and no error will be generated):
```py
from asyncio_mqtt import Client


client = Client("192.168.5.24", username="admin", password="123456")

async def run3(message):
    await client.publish("humidity/outside", payload=message)


async def gather_task3():
    await client.connect()
    tasks = [asyncio.create_task(run3(i)) for i in range(10)]
    await asyncio.gather(*tasks)
    await client.disconnect()

if __name__ == '__main__':
    import asyncio
    asyncio.get_event_loop().run_until_complete(gather_task3())
```

So is it possible to consider from other angles and then keep connect/disconnect.
We can not recommend manually calling connect/disconnect unless you understand the logic behind it.

frederikaalund (approved):

frederikaalund:
First of all, I apologize about the long delay.

In any case, this PR now looks good to me. 👍 Though I can't merge it in because GitHub says "This branch cannot be rebased due to conflicts". I guess that the main branch updated (probably a bot-based commit) in the meantime.

@vvanglro Can I get you to rebase this PR on top of the master branch? Thanks! Then I'll merge it in right away. :)

vvanglro:
I can't reproduce the error in CI locally. This seems to be a problem with mypy?

I use python3.8 locally, and python3.10 in CI throws an error.

image

empicano:
The CI tests also fail for `3.8`, see https://github.com/sbtinstruments/asyncio-mqtt/actions/runs/5217505877/jobs/9431899291?pr=216; pre-commit doesn't run the unit tests, you can run the tests by running `pytest` 🙂

frederikaalund:
It was just a simple merge conflict. I added a merge commit on it to fix it. :)

Thank you for a great contribution to asyncio-mqtt. 👍

thalesmaoa:
This is on the main branch now? Do I still need to call disconnect?

frederikaalund:
> This is on the main branch now?

Yes, and part of the v1.0.0 release. :)

>Do I still need to call disconnect?

We did not remove `disconnect` (too big of a change). You may still call `connect/disconnect` instead of using the context manager approach.

PR #217, opened 2023-05-18
Fix Ruff and typing errors

JonathanPlasse:


JonathanPlasse:
Current dependencies on/for this PR:
* main
  * **PR #217** <a href="https://app.graphite.dev/github/pr/sbtinstruments/asyncio-mqtt/217" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sbtinstruments/asyncio-mqtt/217?utm_source=stack-comment).

asyncio_mqtt/client.py
```diff
@@ -78,7 +78,7 @@ class TLSParameters:
 
 # Proxy parameters class
 class ProxySettings:
-    def __init__(
+    def __init__(  # noqa: PLR0913
```
JonathanPlasse, line 81: The `PLR0913` was updated to take all arguments into account.

asyncio_mqtt/client.py
```diff
@@ -99,10 +99,7 @@ def __init__(
 # See the overloads of `socket.setsockopt` for details.
 SocketOption: TypeAlias = "tuple[int, int, int | bytes] | tuple[int, int, None, int]"
 
-SubscribeTopic: TypeAlias = (
-    "str | tuple[str, mqtt.SubscribeOptions] | list[tuple[str, mqtt.SubscribeOptions]]"
-    " | list[tuple[str, int]]"
-)
+SubscribeTopic: TypeAlias = "str | tuple[str, mqtt.SubscribeOptions] | list[tuple[str, mqtt.SubscribeOptions]] | list[tuple[str, int]]"
```
JonathanPlasse, line 102: Fix multi-line type alias, Pylance/Pyright did not like it.

asyncio_mqtt/error.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, rc: int | mqtt.ReasonCodes | None, *args: Any) -> None:
 
     def __str__(self) -> str:
         if isinstance(self.rc, mqtt.ReasonCodes):
-            return f"[code:{self.rc.value}] {str(self.rc)}"
+            return f"[code:{self.rc.value}] {self.rc!s}"
```
JonathanPlasse, line 20: New rule `RUF010` that uses f-string conversion instead of a call to str.

pyproject.toml
```diff
@@ -41,7 +41,7 @@ dynamic = ["version"]
 "Issue tracker" = "https://github.com/sbtinstruments/asyncio-mqtt/issues"
 
 [project.optional-dependencies]
-lint = ["mypy>=1.2.0", "ruff==0.0.265", "types-paho-mqtt>=1.6.0.6"]
+lint = ["mypy>=1.3.0", "ruff==0.0.267", "types-paho-mqtt>=1.6.0.6"]
```
JonathanPlasse, line 44: Update dependencies to be the same as pre-commit.

pyproject.toml
```diff
@@ -66,7 +66,7 @@ color_output = true
 profile = "black"
 
 [tool.ruff]
-required-version = "0.0.265"
+required-version = "0.0.267"
```
JonathanPlasse, line 69: `required-version` must be kept in sync with the `pre-commit` version so as to not break code when Ruff is updated.

asyncio_mqtt/client.py
JonathanPlasse: @empicano

empicano (approved):
I'm a bit torn with enabling _all_ of `ruff`'s rules. In other projects, I try to minimize the rules to those few where I don't tolerate exceptions -- something like `E,W,F,I,N`. I feel that if it's easier to satisfy the rules, there's less chance that people just start ignoring them completely after a while (or start getting frustrated when you enforce them).

Maybe this could make it a bit easier to contribute here as well. I'm not sure every contributor is confident enough to add a `noqa` and actively break a rule that we set up.

Just my two cents. Let me know what you think, in any case, I'm also fine with the status quo 🙂

JonathanPlasse:
Totally open to explicitly which rules we want.
Using `ALL` is discouraged.

JonathanPlasse:
For me, it is ok to have a PR that does not pass Ruff and the maintainer can help fix the different errors.

frederikaalund (approved):
LGTM. 👍 Thank you for the maintenance work. :)

#### On the topic of how many rules to enable

In the repos that I fully control, I favor to enable *all* the rules per default and then disable specific rules on a case-by-case basis. I'm the `-Wpedantic` guy on the C++ team. 😅 

I'm not so familiar with ruff so I can't tell if the current `select` list is "near to `ALL` but with few exceptions" atm. @JonathanPlasse, is it?

I agree with @empicano that that kind of approach is a contribution barrier. So for a public library like asyncio-mqtt we should be more lenient. I.e., to encourage third party contributions.

I like @JonathanPlasse's idea to enable the strict rules and then help contributors fix them (even do it ourselves). That's a compromise between "strict code quality requirements" an "encourage contribution".

JonathanPlasse:
> In the repos that I fully control, I favor to enable all the rules per default and then disable specific rules on a case-by-case basis. I'm the -Wpedantic guy on the C++ team. sweat_smile

Me too :smile: 

> I'm not so familiar with ruff so I can't tell if the current select list is "near to ALL but with few exceptions" atm. @JonathanPlasse, is it?

Yes, the select is `ALL` with few exceptions.

> I like @JonathanPlasse's idea to enable the strict rules and then help contributors fix them (even do it ourselves). That's a compromise between "strict code quality requirements" an "encourage contribution".

I think we should update `CONTRIBUTING.md` to explain this.

JonathanPlasse:
I added `flake8-todos` with Ruff.
Now it will force you to add the name of the author of the todo.
Like this, no need to do a git blame and it is more robust and faster.
The author of the todo is not necessarily the one that has to solve it.

Issue #109, opened 2022-03-14
QOS1 first incoming message if missed after starting the client


Kopeika:
Debian, Python 3.10, mosquitto as a broker

started:
mosquitto_sub --topic  /mytopic -q 1 -i myclient -c

send one message:
mosquitto_pub -f current/cache_push.json --topic  /mytopic -q 1

it will be received by mosquito_sub

now I stopped mosquitio_sub
and send one message using 
mosquitto_pub -f current/cache_push.json --topic  /mytopic -q 1

it was stored in MQTT broker

when i started 
mosquitto_sub --topic  /mytopic -q 1 -i myclient -c

I got that message.

This is ok. It must work this way.

Now i will use your updated example code instead of mosquito_sub as a client

1) i wil execute it once - so it will store client_id, and subscription information in the MQTT broker
2) i will stop client
3) i wil send message using mosquitto_pub -f current/cache_push.json --topic  /mytopic -q 1
4) i will start client 
5) no incoming messages

if i will send 3 messages instead of one at 3) - i will receive two of them. 

What am I doing wrong?

Thank you!

`
import asyncio
import logging
from asyncio_mqtt import Client, MqttError

logger = logging.getLogger(__name__)

async def tick():
    while True:
        #logger.info("Tick")
        await asyncio.sleep(1)

async def test():
    while True:
        try:
            logger.info("Connecting to MQTT")
            async with Client("localhost", client_id="minmax" , clean_session=False) as client:
                logger.info("Connection to MQTT open")
                async with client.unfiltered_messages() as messages:
                    await client.subscribe("#",1)
                    async for message in messages:
                        logger.info(
                            "Message %s %s", message.topic, message.payload.decode()
                        )
                # await asyncio.sleep(2)
        except MqttError as e:
            logger.error("Connection to MQTT closed: " + str(e))
        except Exception:
            logger.exception("Connection to MQTT closed")
        await asyncio.sleep(3)

def main():
    logging.basicConfig(
        level=logging.DEBUG,
        format="%(asctime)s %(levelname)s %(message)s",
        datefmt="%Y-%m-%d %H:%M:%S",
    )
    asyncio.get_event_loop().run_until_complete(asyncio.wait([test(), tick()]))

if __name__ == "__main__":
    main()

`

frederikaalund:
Hi and thanks for opening this issue. Let me have a look. :)

[...]

PR #220, opened 2023-05-27
feat: change message_queue to class attribute

vvanglro:
1.  Change message_queue to class attribute.
2. Add message queue test case.
3. Change custom message docs.

### DIFF
--- a/asyncio_mqtt/client.py
+++ b/asyncio_mqtt/client.py
@@ -269,6 +269,10 @@ class Client:
         websocket_headers: WebSocketHeaders | None = None,
         max_inflight_messages: int | None = None,
         max_queued_messages: int | None = None,
+        message_queue_class: type[
+            asyncio.Queue[Message | mqtt.MQTTMessage]
+        ] = asyncio.Queue,
+        message_queue_maxsize: int = 0,
     ) -> None:
         self._hostname = hostname
         self._port = port
@@ -367,6 +371,9 @@ class Client:
         if socket_options is None:
             socket_options = ()
         self._socket_options = tuple(socket_options)
+        self._message_queue_class: asyncio.Queue[
+            Message | mqtt.MQTTMessage
+        ] = message_queue_class(message_queue_maxsize)
 
     @property
     def id(  # noqa: A003 # TODO(jonathan): When doing BREAKING CHANGES rename to avoid shadowing builtin id
@@ -498,7 +505,7 @@ class Client:
 
     @asynccontextmanager
     async def filtered_messages(
-        self, topic_filter: str, *, queue_maxsize: int = 0
+        self, topic_filter: str
     ) -> AsyncGenerator[AsyncGenerator[mqtt.MQTTMessage, None], None]:
         """Return async generator of messages that match the given filter."""
         self._logger.warning(
@@ -506,7 +513,7 @@ class Client:
             " Use messages() together with Topic.matches() instead."
         )
         callback, generator = self._deprecated_callback_and_generator(
-            log_context=f'topic_filter="{topic_filter}"', queue_maxsize=queue_maxsize
+            log_context=f'topic_filter="{topic_filter}"'
         )
         try:
             self._client.message_callback_add(topic_filter, callback)
@@ -518,7 +525,7 @@ class Client:
 
     @asynccontextmanager
     async def unfiltered_messages(
-        self, *, queue_maxsize: int = 0
+        self,
     ) -> AsyncGenerator[AsyncGenerator[mqtt.MQTTMessage, None], None]:
         """Return async generator of all messages that are not caught in filters."""
         self._logger.warning(
@@ -530,7 +537,7 @@ class Client:
             msg = "Only a single unfiltered_messages generator can be used at a time."
             raise RuntimeError(msg)
         callback, generator = self._deprecated_callback_and_generator(
-            log_context="unfiltered", queue_maxsize=queue_maxsize
+            log_context="unfiltered"
         )
         try:
             self._unfiltered_messages_callback = callback
@@ -543,9 +550,6 @@ class Client:
     @asynccontextmanager
     async def messages(
         self,
-        *,
-        queue_class: type[asyncio.Queue[Message]] = asyncio.Queue,
-        queue_maxsize: int = 0,
     ) -> AsyncGenerator[AsyncGenerator[Message, None], None]:
         """Return async generator of incoming messages.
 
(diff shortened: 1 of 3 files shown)


empicano:
Hi vvanglro, thanks for all these contributions, very cool! 😎

I only had a quick glance over this, so I have no feedback on the code. Re:docs the new parameters to the client are missing in `connecting-to-the-broker.md`.

frederikaalund:
Thanks for yet another contribution to asyncio-mqtt. 👍 Keep it up! 💪 :)

I like the idea that the user can choose a `Client`-wide default for the message queue class. I also like the current approach where the user can choose a queue class per message loop (i.e., per call to `messages`). Let's keep both!

I propose that we:
 * Add `message_queue_class` to `Client.__init__` (this PR already does so)
 * Simply save `self._message_queue_class = message_queue_class` *and not an instance of said class* (this PR does the latter, which I think is a bug anyhow).
 * Keep the `def messages(..., queue_class)` kwarg as is (this PR currently removes this kwag).
 * Simply use `queue_class` if given and if not given, fall back to `self._message_queue_class`

I thank that gives of the best of both worlds: A client-wide default that the user can override per message loop. :) Does it make sense?

On a side note: I really like the test that you have in place for this! 👍 It's a great example of how to use this feature.

Thanks again for you contribution to asyncio-mqtt!

vvanglro:
@frederikaalund Sorry, maybe I didn't explain the reason for this change.


This is multiple messages, and different message queues, But no matter which topic the message belongs to, the message will be put in all message queues, So in this case, I don't think it is necessary to set up a queue for each message.

`def _on_message`Every time there is a message, all the callback methods will be called back to put the message in the queue.
https://github.com/sbtinstruments/asyncio-mqtt/blob/9366b32b6ea95dc9dc95b23c71672d17bd5d296f/asyncio_mqtt/client.py#L816-L818
```py
import asyncio
import asyncio_mqtt as aiomqtt


class CustomPriorityQueue(asyncio.PriorityQueue):
    def _put(self, item):
        priority = 2
        if item.topic.matches("humidity/#"):  # Assign priority
            priority = 1
        super()._put((priority, item))

    def _get(self):
        return super()._get()[1]


async def message1(client):
    async with client.messages(queue_class=CustomPriorityQueue) as messages:
            await asyncio.sleep(0.1)
            print("temperature/#:", message.payload)


async def message2(client):
    async with client.messages(queue_class=asyncio.Queue) as messages:
            await asyncio.sleep(0.1)
            print("humidity/#:", message.payload)


async def main():
        task = [asyncio.create_task(message1(client)), asyncio.create_task(message2(client))]
        await asyncio.gather(*task)


asyncio.run(main())
```


```py
from asyncio_mqtt import Client
client = Client("test.mosquitto.org")


async def run():

    await client.connect()
    await client.publish("temperature/outside", payload=1.312)
    await client.publish("humidity/outside", payload=21)
    await client.disconnect()


if __name__ == '__main__':
    import asyncio
    asyncio.get_event_loop().run_until_complete(run())
```

These are my thoughts, I don't know if I missed something else.

frederikaalund:
Thanks for the detailed response! 👍 Now I understand what the purpose of this PR is. I agree that it makes sense to use a client-wide "queue" for all incoming messages. I'm worried about the implementation in this PR. I think the client-wide queue opens us up for race conditions. Take this counter example for instance:

```python
import asyncio
import asyncio_mqtt as aiomqtt


async def fast_consumer(client: aiomqtt.Client):
        await client.subscribe("counter_example/#")
            print(f"Fast consumer received: {message.payload}")

async def slow_consumer(client: aiomqtt.Client):
        await client.subscribe("counter_example/#")
            await asyncio.sleep(1)
            print(f"Slow consumer received: {message.payload}")


async def producer(client: aiomqtt.Client):
    for i in range(5):
        await client.publish("counter_example/int", i)
        await asyncio.sleep(0.1)

async def main():
        tasks = [
            asyncio.create_task(fast_consumer(client)),
            asyncio.create_task(slow_consumer(client)),
            asyncio.create_task(producer(client))]
        await asyncio.gather(*tasks)


asyncio.run(main())
```

Normally (without this PR), the counter example code always produce:

```shell
Fast consumer received: b'0'
Fast consumer received: b'1'
Fast consumer received: b'2'
Fast consumer received: b'3'
Fast consumer received: b'4'
Slow consumer received: b'0'
Slow consumer received: b'1'
Slow consumer received: b'2'
Slow consumer received: b'3'
Slow consumer received: b'4'
```

With this PR, the counter example code produces something like:

```shell
(asyncio-mqtt-py3.10) fpa@sbt-fpa-desktop1:/projects/asyncio-mqtt$ poetry run python ./counter_example.py 
Fast consumer received: b'0'
Fast consumer received: b'1'
Fast consumer received: b'1'
Fast consumer received: b'2'
Fast consumer received: b'2'
Fast consumer received: b'3'
Fast consumer received: b'3'
Fast consumer received: b'4'
Fast consumer received: b'4'
Slow consumer received: b'0'
```

Note how the fast consumer accidentally consumed almost every message! That's bad.

I hope that makes sense. Let me know if I missed something. :)

vvanglro:
Oh, I understand🤯

empicano:
I think there's a good idea here. If we go all in on _one_ client-wide queue, we could simplify:

```python
            print(message.payload)
```

to something like:


```python
        print(message.payload)
```

That could be neat. At first glance, it's also maybe not super clear that each `client.messages()` creates a separate queue.

This makes the default use case simpler (And avoids issues like #109) but breaks the use case where we have:

- fast and slow consumers that handle the same messages
- consumers that should handle the same messages in a different order

(Those are the ones I can think of, at least.)

However, if someone needs this functionality, it should be possible to implement this on top by transferring the messages to other queues inside `async for message in client.messages:`.

What are your thoughts on that?

frederikaalund:
Personally, I like this "single client-wide queue" approach. I agree that it simplifies things (which I guess is why @vvanglro made this PR to begin with).

Pros:
 * Simpler implementation and API (this PR shows how and @empicano's comment details further)
 * More intuitive since we avoid that whole "multiple queues" thing. I don't know if our users actually realize that's how it works.

Cons:
 * Breaking change. This is v2.0.0 material. :)
 
Maybe we should open a GitHub discussion and collect all these ideas there?

Again, thanks for @vvanglro for putting this PR out there and opening the discussion. 👍

empicano:
Good idea 👍 I opened one [here](https://github.com/sbtinstruments/aiomqtt/discussions/226)

PR #225, opened 2023-06-14
feat: match topic before message enters queue

vvanglro:
Add a topic optional parameter to messages, so that each queue only receives messages of the corresponding topic.

This is the case where topic is not used, it is the same as before:
```py
import asyncio
import aiomqtt


async def fast_consumer(client: aiomqtt.Client):
        await client.subscribe("counter_example/#")
            print(f"Fast consumer received: {message.payload}")


async def slow_consumer(client: aiomqtt.Client):
            await client.subscribe("qwer/#")
                print(f"Slow consumer received: {message.payload}")


async def producer(client: aiomqtt.Client):
    await client.publish("counter_example/int", "counter_example")
    await client.publish("qwer/int", "qwer")


async def main():
        tasks = [
            asyncio.create_task(fast_consumer(client)),
            asyncio.create_task(slow_consumer(client)),
            asyncio.create_task(producer(client))]
        await asyncio.gather(*tasks)


asyncio.run(main())

```
Output:
```sh
Fast consumer received: b'counter_example'
Slow consumer received: b'counter_example'
Fast consumer received: b'qwer'
Slow consumer received: b'qwer'
```

This is the case where topic is used:
```py
import asyncio
import aiomqtt


async def fast_consumer(client: aiomqtt.Client):
    async with client.messages(topic="counter_example/#") as messages:
        await client.subscribe("counter_example/#")
            print(f"Fast consumer received: {message.payload}")


async def slow_consumer(client: aiomqtt.Client):
        async with client.messages(topic="qwer/#") as messages:
            await client.subscribe("qwer/#")
                print(f"Slow consumer received: {message.payload}")


async def producer(client: aiomqtt.Client):
    await client.publish("counter_example/int", "counter_example")
    await client.publish("qwer/int", "qwer")


async def main():
        tasks = [
            asyncio.create_task(fast_consumer(client)),
            asyncio.create_task(slow_consumer(client)),
            asyncio.create_task(producer(client))]
        await asyncio.gather(*tasks)


asyncio.run(main())
```
Output:
```sh
Fast consumer received: b'counter_example'
Slow consumer received: b'qwer'
```

I thought it would be better than matching after receiving the message before.

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -571,6 +571,7 @@ class Client:
         *,
         queue_class: type[asyncio.Queue[Message]] = asyncio.Queue,
         queue_maxsize: int = 0,
+        topic: str = None,
     ) -> AsyncGenerator[AsyncGenerator[Message, None], None]:
         """Return async generator of incoming messages.
 
@@ -579,7 +580,7 @@ class Client:
         If queue_maxsize is less than or equal to zero, the queue size is infinite.
         """
         callback, generator = self._callback_and_generator(
-            queue_class=queue_class, queue_maxsize=queue_maxsize
+            queue_class=queue_class, queue_maxsize=queue_maxsize, topic=topic
         )
         try:
             # Add to the list of callbacks to call when a message is received
@@ -646,6 +647,7 @@ class Client:
         *,
         queue_class: type[asyncio.Queue[Message]] = asyncio.Queue,
         queue_maxsize: int = 0,
+        topic: str = None,
     ) -> tuple[Callable[[Message], None], AsyncGenerator[Message, None]]:
         # Queue to hold the incoming messages
         messages: asyncio.Queue[Message] = queue_class(maxsize=queue_maxsize)
@@ -653,7 +655,11 @@ class Client:
         def _callback(message: Message) -> None:
             """Put the new message in the queue."""
             try:
-                messages.put_nowait(message)
+                if topic:
+                    if message.topic.matches(topic):
+                        messages.put_nowait(message)
+                else:
+                    messages.put_nowait(message)
             except asyncio.QueueFull:
                 self._logger.warning("Message queue is full. Discarding message.")
 
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -571,7 +571,7 @@ class Client:
         *,
         queue_class: type[asyncio.Queue[Message]] = asyncio.Queue,
         queue_maxsize: int = 0,
-        topic: str = None,
+        topic: str | None = None,
     ) -> AsyncGenerator[AsyncGenerator[Message, None], None]:
         """Return async generator of incoming messages.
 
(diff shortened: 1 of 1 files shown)


empicano:
Wow, we're getting a lot of contributions from you, that's really cool! 😎

So this lets us write:

```python
async with client.messages(topic="temperature/inside") as messages:
        print(message.payload)
```

instead of:

```python
        if message.topic.matches("temperature/inside"):
            print(message.payload)
```

Saves a line, which is always neat 😄 On the other hand, I'm a fan of the line "There should be one -- and preferably only one -- obvious way to do it" in [The Zen of Python](https://peps.python.org/pep-0020/).

The new `topic` parameter is only useful when we use multiple queues and other than saving one line, it's not really different from using `message.topic.matches`, which is a lot more flexible as well. One advantage that I _do_ see is that the queues have fewer entries.

This is slightly related to [my comment in #220](https://github.com/sbtinstruments/aiomqtt/pull/220#issuecomment-1592041102), I think I'd like to solve use cases like this by sorting the messages into other queues inside `async for message in messages:` instead of implementing this inside aiomqtt.

Happy to hear other opinions 🙂

frederikaalund:
Thanks for all these PRs! It's great to see all these contributions from you. 👍 

I second @empicano thoughts here. In fact, we already have the deprecated [`Client.filtered_messages`](https://github.com/sbtinstruments/aiomqtt/blob/main/aiomqtt/client.py#L525) that does exactly this! 😅 We deprecated this in favor of the approach that @empicano lines out in his comment (it does the same thing but is more flexible).

I too am more in favor of the "single client-wide queue" approach that #220 proposes. In any case, I welcome the discussion! Speaking of that, now that we are officially "1.0.0", it's a good time to start the discussion on "2.0.0" already. 👍

@vvanglro what do you think? Did you consider the `message.topic.matches` approach? If so, how do you see the pros/cons between this PR and the `message.topic.matches` approach and the existing (and deprecated) `Client.filtered_messages`? :)

vvanglro:
First of all, my following views are based on the multi-process situation of web services:

- About client-scoped queues:
  1. This will use less memory as there won't be multiple queues.
  2. Regarding the competition problem, I think it is meaningless for a single client to subscribe to multiple same topics.
  3. mqtt itself is in broadcast mode, so in multi-process mode, there will be a problem of repeated consumption. In this case, the shared subscription brought by 5.0 can be used. (Probably not related to range queues)
  
- About matching messages:
  1. If there are multiple queues at present, I think there is still a difference between this PR and the matching after getting the message. Judging whether it matches when entering the queue can reduce the length of the queue and ensure that each message queue only receives messages of the corresponding topic.
  2. If it is a client-wide queue, then all subscribed topic messages will enter this queue, so after getting the message, match different topic messages, or we can pass in the topic to be processed in client.messages(topic="xxx"), Then only return this topic message when iterating messages(Reduce redundancy caused by if-else).

PR #231, opened 2023-07-08
Add reference section to docs

empicano:
I added Google-style docstrings to most of the classes to autogenerate a reference section in the docs (+ some smaller documentation tweaks). You can preview the docs as usual with `./scripts/setup && ./scripts/docs --reload`. Happy to hear what you think!

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,11 +5,13 @@ on:
     branches: [main]
     paths:
       - "docs/**"
+      - "aiomqtt/**"
       - README.md
       - .github/workflows/docs.yml
   pull_request:
     paths:
       - "docs/**"
+      - "aiomqtt/**"
       - README.md
       - .github/workflows/docs.yml
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,9 @@ We're very happy about contributions to aiomqtt! ✨
 
 ## The documentation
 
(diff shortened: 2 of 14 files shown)


frederikaalund:
I like it. 😄👍 Think it'll make it easier to approach aiomqtt for newcomers. Good idea it document the function arguments directly in the docstrings. I know that some editors/IDEs (e.g., VS Code) use the docstrings to display a help message to the user. This will aid them. :)

I also like the explanations about context managers (contrasted to connect/disconnect). This is probably what confuses most people about aiomqtt. Good that you mentioned the workarounds (direct calls to aenter/aexit) in there. That enables our users to use the old-style approach but their linter will shout at them if they have one (because they call dunder methods). :)

Thanks for yet another great contribution to aiomqtt! 👍

JonathanPlasse:
You should update [`ruff.pydocstyle.convention`](https://beta.ruff.rs/docs/settings/#pydocstyle-convention) in `pyproject.toml` to `google`.
By doing this it found 3 rule violations:
```console
aiomqtt/client.py:540:15: D417 Missing argument descriptions in the docstring: `**kwargs`, `*args`, `properties`
aiomqtt/client.py:575:15: D417 Missing argument descriptions in the docstring: `**kwargs`, `*args`
aiomqtt/client.py:602:15: D417 Missing argument descriptions in the docstring: `**kwargs`, `*args`
Found 3 errors.
```

empicano:
Wow, that's cool that ruff can check that. Great tip! 👍

Issue #224, opened 2023-06-12
Race condition in unsubscribe test workflow


empicano:
There seems to be a race condition in the [test_client_unsubscribe](https://github.com/sbtinstruments/asyncio-mqtt/blob/54429028b8603e96b47341fa327a24e0d42b26e9/tests/test_client.py#L153) test when it runs in the test matrix. This surfaced with the switch to poetry, maybe I missed something there as well.

When I run the workflows one after the other everything is fine, but when they run at the same time [they often fail](https://github.com/sbtinstruments/asyncio-mqtt/actions/runs/5246555751/jobs/9475687465).

To mitigate this we could spin up our own mosquitto broker via Docker for each test run instead of using the public "test.mosquitto.org" one.

frederikaalund:
It does sound "race condition"-like since it only happens sometimes.

> Maybe this could be mitigated by spinning up our own mosquitto broker via Docker for each test run instead of using the public "test.mosquitto.org" one.

That's a great idea. 👍 Would make the tests more resilient to third party disturbances (e.g., if test.mosquitto.org is overloaded).

JonathanPlasse:
I do not know if we can use Docker when testing on Windows.

frederikaalund:
I just got bit by this! 😄

I think I found the culprit though: The tests all ran as python 3.10. In turn, that caused the `TOPIC_HEADER` to contain `3.10` for all our tests. In turn, our test suite published to the *same topic in parallel*. That topic-based race condition triggered in `test_client_unsubscribe`.

I fixed it by adding these lines to the test workflow:
```
      - name: Make Poetry use the active python version 
        run: poetry config virtualenvs.prefer-active-python true
```

It makes poetry use the python version that we just installed. With this in place, I can see that the tests actually execute with the respective python versions (as they should). This fixes the race condition since `TOPIC_HEADER` is unique again.

---

I still think it's a good idea to use something else than test.mosquitto.org for this. :) Let's keep this issue open for that.

empicano:
That's good detective work 😎 I totally missed that!

The change to the setup script broke the docs workflow, hehe. I tried around with the workflows with your insight in mind and noticed that the problem stems from the order that the `setup-python` and `install-poetry` actions are run.

[...]

PR #232, opened 2023-07-11
don't suppress exception on aexit

edenhaus:
With #216 the "early out on disconnected" code was move from the function `__aexit__` to `disconnect`. The `disconnect` function is called inside a try/except block in the function `__aexit__`, which suppresses the disconnected exception. This results in that the context manager exits without exception also when there was a disconnect-exception.

Example:
- start mqtt broker
- run a similar code:
```python
async with Client("localhost") as client:
            print(message.payload)
```
- shutdown broker

expection: I get some sort of exception
actual: no exception, only a log entry.

This bug was found on the upgrade from `0.16.1` to `1.0.0`. Afterwards the following test was failing https://github.com/DeebotUniverse/client.py/blob/3d7467e3dcf338e9d848dd5fb3ad32a5ee7fab8e/tests/test_mqtt_client.py#L76C1-L76C1
The test test the described example above.

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -514,7 +514,8 @@ class Client:
     def _early_out_on_disconnected(self)-> bool:
          # Early out if already disconnected...
         if self._disconnected.done():
-            if (disc_exc:= self._disconnected.exception()) is not None:
+            disc_exc = self._disconnected.exception()
+            if disc_exc is not None:
                 # ...by raising the error that caused the disconnect
                 raise disc_exc
             # ...by returning since the disconnect was intentional
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -511,15 +511,19 @@ class Client:
         if self._disconnected.done():
             self._disconnected = asyncio.Future()
 
-    async def disconnect(self, *, timeout: float | None = None) -> None:
-        """Disconnect from the broker."""
-        # Early out if already disconnected...
+    def _early_out_on_disconnected(self)-> bool:
+         # Early out if already disconnected...
         if self._disconnected.done():
-            disc_exc = self._disconnected.exception()
-            if disc_exc is not None:
+            if (disc_exc:= self._disconnected.exception()) is not None:
                 # ...by raising the error that caused the disconnect
                 raise disc_exc
             # ...by returning since the disconnect was intentional
+            return True
+        return False
+
+    async def disconnect(self, *, timeout: float | None = None) -> None:
+        """Disconnect from the broker."""
+        if self._early_out_on_disconnected():
             return
         # Try to gracefully disconnect from the broker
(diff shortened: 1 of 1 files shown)


aiomqtt/client.py
```diff
@@ -511,15 +511,19 @@ async def connect(self, *, timeout: float | None = None) -> None:
...
 
-    async def disconnect(self, *, timeout: float | None = None) -> None:
-        """Disconnect from the broker."""
-        # Early out if already disconnected...
+    def _early_out_on_disconnected(self)-> bool:
+         # Early out if already disconnected...
         if self._disconnected.done():
-            disc_exc = self._disconnected.exception()
-            if disc_exc is not None:
+            if (disc_exc:= self._disconnected.exception()) is not None:
```
frederikaalund, line 517: Rewrite to avoid `:=` since that's a python 3.8 feature (and we still support 3.7).

...Good use case for `:=`, though. :)

frederikaalund (changes requested):
Hi Robert, thanks for opening this pull request. 👍  Let me have a look. :)

You're right. This seems like a regression. Indeed, the client should raise (the Exception that caused a disconnect) on exit. The current implementation erroneously suppresses (and logs) this disconnect exception. That's my bad; I missed it in the code review. Good thing that you had a test case for it. If you don't mind, could you implement a similar test case here in aiomqtt? 👍 

In any case, I like this PR overall. I only have requests for minor changes. See my review for details. Let me know if you have any questions. Also, you need to fix the formatting/style issues (see the CI runs for details).

Thanks again for opening this pull request.

Best regards,
~Frederik

edenhaus:
For tests I would wait until https://github.com/guillotinaweb/pytest-docker-fixtures/pull/33 is merged, than we have a moquitto docker container for tests. This would also fix #224

Are you fine to merge it without tests? So I can upgrade my lib to the new version and get rid of the log message about the rename

empicano:
Hi Robert, I'll ping @frederikaalund here; Maybe this got lost 😉

As a quick fix to getting rid of the log message about the rename you could restrict the version of asyncio-mqtt to `<= 0.16.1`. Version `0.16.1` is exactly the same as `0.16.2` apart from the log message.

frederikaalund:
Thanks for the ping, @empicano. Been a busy week at work due to technical issues. :)

In any case, this PR now looks good to me. :+1: Sorry about the delay in my response.

> For tests I would wait until https://github.com/guillotinaweb/pytest-docker-fixtures/pull/33 is merged, than we have a moquitto docker container for tests. This would also fix https://github.com/sbtinstruments/aiomqtt/issues/224
>
> Are you fine to merge it without tests? So I can upgrade my lib to the new version and get rid of the log message about the rename

Yes, that's completely fine with me.

Thank you for your contribution to aiomqtt.

Issue #244, opened 2023-08-18
Reconnection example throws "Does not support reentrant"


westbywest:
Adapting the [reconnection example](https://sbtinstruments.github.io/aiomqtt/reconnection.html) to throw exceptions verbosely:
```
import asyncio
import aiomqtt

async def main():
    client = aiomqtt.Client("testbad.mosquitto.org")
    interval = 5  # Seconds
    while True:
        try:
            async with client:
                async with client.messages() as messages:
                    await client.subscribe("humidity/#")
                    async for message in messages:
                        print(message.payload)
        except aiomqtt.MqttError as e:
            print(f"MQTT error {e}; Reconnecting in {interval} seconds ...")
            await asyncio.sleep(interval)

asyncio.run(main())
```
Subsequent reconnection attempts actually will always fail on "Does not support reentrant," even if the target host becomes available.
```
MQTT error [Errno -2] Name or service not known; Reconnecting in 5 seconds ...
MQTT error Does not support reentrant; Reconnecting in 5 seconds ...
MQTT error Does not support reentrant; Reconnecting in 5 seconds ...
```
For reconnection to succeed, I have to move `client = ...` statement to within the while loop.  Is this intended usage, and the example happens to be incorrect?

Observed under python 3.10.7, asyncio 3.4.3, aiomqtt 1.1.0.

vvanglro:
@westbywest Hi, I can't reproduce it. I use the same code. After restarting the mqtt service. It can still be reconnected normally.

empicano:
Hi Ben,

Thanks for the minimal example, I can reproduce it on my side 👍 Moving the `client` statement inside the while loop is a good option in the meantime, but we _do_ want it to work like in the example (without re-initializing the client instance), so it seems there's something broken, good catch!

For reference, we want the client to be [reusable, but not reentrant](https://docs.python.org/3/library/contextlib.html#reusable-context-managers), the corresponding PR is #216 (by @vvanglro, thanks again! 🚀).

This is an interesting one! I tested this a bit more thouroughly with a local Docker mosquitto broker that I shut down. With some logging on the exact exception types and on the [relevant lock](https://github.com/sbtinstruments/aiomqtt/blob/b3c3de39b7752151f761028c37766b6915aeb436/aiomqtt/client.py#L465), this gives:

```
lock acquired
lock released
<class 'aiomqtt.error.MqttCodeError'>
[...]

PR #245, opened 2023-08-21
fix: client aenter connect error lock release

vvanglro:
fix: #244


vvanglro:
https://github.com/snok/install-poetry/issues/94

aiomqtt/client.py
```diff
@@ -1036,7 +1036,11 @@ async def __aenter__(self) -> Client:
             msg = "Does not support reentrant"
             raise MqttReentrantError(msg)
         await self._lock.acquire()
-        await self.connect()
+        try:
+            await self.connect()
+        except Exception:
+            self._lock.release()
+            raise
```
empicano, line 1043: Very nice 👍

tests/test_client.py
```diff
@@ -536,3 +536,12 @@ async def test_client_connecting_disconnected_done() -> None:
     client._disconnected.set_result(None)
     await client.connect()
     await client.disconnect()
+
+
+@pytest.mark.network
+async def test_client_aenter_connect_error_lock_release() -> None:
+    client = Client(hostname="aenter_connect_error_lock_release")
+    with pytest.raises(MqttError):  # noqa: PT012
+        async with client:
```
empicano, line 545: You could use `await client.__aenter__()` here instead, this way we do not have to ignore the rule. Otherwise, good test! 👍

empicano:
Hi there,

Great that you put work into this, thank you! 😎 I believe the reason that the tests are failing is that the [poetry dropped support for Python 3.7 in 1.6.0](https://github.com/python-poetry/poetry/releases/tag/1.6.0) rather than the issue you linked. See also your first commit, where all those workflows that use Python 3.7 fail, but not the others.

Python 3.7 reached its end of life, so I think we can drop it anyways. I opened PR #246 to do that.

Your PR looks good to me otherwise 🙂

vvanglro:
> Python 3.7 reached its end of life, so I think we can drop it anyways. I opened PR #246 to do that.

When your PR is merged, I will merge it once.

frederikaalund:
Looks good to me as well. :+1:  I just reviewed #246 as well.

empicano:
Thanks for another great contribution to aiomqtt, @vvanglro! 💪

PR #246, opened 2023-08-22
Drop support for Python 3.7

empicano:
Python 3.7 has reached its [end-of-life](https://devguide.python.org/versions/). Poetry dropped support in 1.6.0 so our test workflow is breaking. @vvanglro made an ad-hoc fix in #245 for that, but I think we're good to go to drop it as well in the next release.

Is there anything we can simplify in the code with this step that I missed?


aiomqtt/client.py
```diff
@@ -549,6 +549,7 @@ def _force_disconnect(self) -> None:
     @_outgoing_call
     async def subscribe(  # noqa: PLR0913
         self,
+        /,
```
frederikaalund, line 552: This is technically a breaking change.

This goes for all public methods (including `subscribe`, `unsubscribe`, `publish`).

Before this PR, our users could do:

```python
client.subscribe(topic="floors/first")
```

After this PR, the code above gives an error.

To avoid breaking stuff, I think that we should postpone this change until a 2.0 release. :)

Does it make sense?

frederikaalund (changes requested):
Good idea to drop Python 3.7 support! I think that we've done our maintenance duty in that part (most other libraries dropped it already). :smile: 

Besides the positional-only arguments, we also got the walrus operator (`:=`) in python 3.8.latter. There may be a use case for that somewhere but it's not really worth the refactor IMHO. :)

See my review for details. Let me know if you have any questions! :+1:

aiomqtt/client.py
empicano: I tested this just now and the example works for me. It fails when I put the `/` _after_ the `topic` parameter (making `topic` positional-only). I've never used this feature before, however, so could be that I made an error somewhere!

I mainly added it because mypy wanted it. I'm ok with ignoring mypy's concerns as well and postponing that change, like you propose 🙂

frederikaalund: You are completely right. :+1:  My bad. I misunderstood the semantics of the `/` operator. :sweat_smile:

frederikaalund (approved):

aiomqtt/client.py
empicano: No worries, thanks for double-checking nonetheless!

JonathanPlasse:
Good work!

PR #247, opened 2023-08-24
refactor: Deprecate connect and disconnect methods

empicano:
As previously discussed in #216, this deprecates calls to `connect` and `disconnect` in favor of `__aenter__` and `__aexit__`.

Hopefully, this will also [simplify the internal state](https://github.com/sbtinstruments/aiomqtt/pull/216#pullrequestreview-1434130633) in the future. On that matter, Frederik, could you elaborate on what you meant by "Have all internal state inside a tagged union (sum type) in rust-style, concurrency-safe semantics."?

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -491,18 +491,16 @@ class Client:
         self._logger.warning(
             "The manual `connect` and `disconnect` methods are deprecated and will be"
             " removed in a future version. The preferred way to connect and disconnect"
-            " the client is to use the context manager interface via `async with`."
-            " In case you're forced to connect and disconnect manually, you can call"
-            " the context manager's `__aenter__` and `__aexit__` methods directly as a"
-            " workaround instead. `__aenter__` is equivalent to `connect`. `__aexit__`"
-            " is equivalent to `disconnect` except that it forces disconnection instead"
-            " of throwing an exception in case the client cannot disconnect normally."
+            " the client is to use the context manager interface via `async with`. In"
+            " case your use case needs to connect and disconnect manually, you can call"
+            " the context manager's `__aenter__` and `__aexit__` methods as an escape"
+            " hatch instead. `__aenter__` is equivalent to `connect`. `__aexit__` is"
+            " equivalent to `disconnect` except that it forces disconnection instead"
+            " of throwing an exception in case the client cannot disconnect cleanly."
             " `__aexit__` expects three arguments: `exc_type`, `exc`, and `tb`. These"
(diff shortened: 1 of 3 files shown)


empicano:
Ready for review now, I pulled in the functionality from `connect` and `disconnect` so that we can simply delete them in the next major version, but thus had to duplicate the tests for the `_disconnect` future handling.

Let me know what you think! 🙂

aiomqtt/client.py
```diff
@@ -1052,21 +1102,30 @@ async def __aexit__(
...
             self._lock.release()
+            if (exc := self._disconnected.exception()) is not None:
+                # If the disconnect wasn't intentional, raise the error that caused it
+                raise exc
+            return
+        # Try to gracefully disconnect from the broker
+        rc = self._client.disconnect()
+        if rc == mqtt.MQTT_ERR_SUCCESS:
+            # Wait for acknowledgement
+            await self._wait_for(self._disconnected, timeout=None)
```
vvanglro, line 1117: If it times out here, the lock will not be released, so my suggestion is to change the release lock to be executed anyway in aexit.

empicano: Thanks for checking the changes! 😊 If we release the lock, we can then call `__aenter__` while the client is (potentially) still connected. I'm not sure what's best here.

In any case, I aimed for having equal logic to before in this PR. I hope that we can simplify the internal state when we remove `connect` and `disconnect`, so that such things are clearer. Right now it's slightly confusing having `_connected`, `_disconnected`, _and_ `_lock` and what these three should specify.

skewty:
Once connected and and `async for message in client.messages` is underway and some unrelated code has a critical error and I want the MQTT connection to shut down and do a graceful exit, what is the best method to accomplish this?

Assuming no new messages are coming in that for loop isn't going to be useful in this situation.

Is there a special / sentinel message I can somehow stuff onto the queue to get the desired result?

empicano:
I did not test this now, but I'd expect the client to do a graceful shutdown when you [cancel the surrounding task](https://docs.python.org/3/library/asyncio-task.html#task-cancellation).

So something like:

```python
async def test():
    with aiomqtt.Client("test.mosquitto.org") as client:
        # Subscribe to topics here
            # Process message

async def main():
    task = asyncio.create_task(test())
    # Do something else
    await asyncio.sleep(1)
    # Oh no, error!
    task.cancel()
    try:
        await task
    except asyncio.CancelledError:
        pass
```

Hope that helps!

PR #249, opened 2023-09-18
release lock only if acquired

spacemanspiff2007:
If connection fails calling `__aexit__` will result in an Lock-exception.
This small check prevents that

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -1066,7 +1066,8 @@ class Client:
                 )
         finally:
             self._force_disconnect()
-            self._lock.release()
+            if self._lock.locked():
+                self._lock.release()
 
 
 def _set_client_socket_defaults(
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -87,6 +87,12 @@ async def test_topic_matches() -> None:
     assert not topic.matches("$test/group/a/b/c")
 
 
+async def test__aexit__():
+    """Test that it's possible to call __aexit__ without __aenter__, e.g. in case of unsucessful __aenter__"""
+    client = Client(HOSTNAME)
+    await client.__aexit__(None, None, None)
+
+
 @pytest.mark.network
 async def test_multiple_messages_generators() -> None:
     """Test that multiple Client.messages() generators can be used at the same time."""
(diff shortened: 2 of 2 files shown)


empicano:
Good catch, and thank you for the PR with a fix! 😊

I wasn't sure if we should allow calling `__aexit__` without a prior `__aexit__` (or an unsuccessful `__aexit__` in this case).

I tested `open()` and `contextlib.ExitStack()`, which don't throw exceptions when calling `__exit__` without `__enter__`. [PEP343](https://peps.python.org/pep-0343/) also states that "[...] `__exit__()` methods should avoid raising errors unless they have actually failed." 👍

Could you add a test case for this? Then we'll merge it straight away 🙂

spacemanspiff2007:
Like this?

> I wasn't sure if we should allow calling `__aexit__` without a prior `__aenter__` (or an unsuccessful `__aenter__` in this case).

Imho it only makes sense, there is nothing to gain if it raises an error.
Also manually calling both methods is the only way to use this library if using a context manager is not an option.

vvanglro:
Using a context manager is the only recommended method and the main direction of subsequent maintenance, because the removal of the connect and disconnect methods is already planned #247.

Based on PEP343, this PR is great.

vvanglro (approved):

spacemanspiff2007:
> Using a context manager is the only recommended method and the main direction of subsequent maintenance, because the removal of the connect and disconnect methods is already planned #247.

I'm aware of that and the docs make this clear.
However I have a central connection manager in my application that manages multiple connection lifecycles.
Using the context manager would mean a complete refactor of the way the application handles connections and loosing features that would have to be implemented in another way..
Instead I just call the context manager methods , e.g. [__aenter__](https://github.com/spacemanspiff2007/HABApp/blob/master/src/HABApp/mqtt/connection/handler.py#L69) or already with a local fix [__aexit__](https://github.com/spacemanspiff2007/HABApp/blob/master/src/HABApp/mqtt/connection/handler.py#L78) and everything works as expected 🤷 
If you have a good idea how I can have feature parity with a context manager I'm happy to hear it. 👍

empicano:
For your use case, using `__aenter__` and `__aexit__` directly sounds like a great escape hatch 👍

The test looks good. Thank you for your contribution to aiomqtt! 😎

PR #263, opened 2024-01-08
release lock only when set

spacemanspiff2007:
I somehow missed this call of release
Follow up #249


JonathanPlasse:
LGTM

empicano:
Thanks for yet another great contribution to aiomqtt! 🚀 😊

PR #271, opened 2024-01-31
Update client.py

### LINKED ISSUE #268: Broken messages generator with client context re-use
Hi, the client context reconnect example in [reconnection.html](https://sbtinstruments.github.io/aiomqtt/reconnection.html) does not work as expected in my testing due to what looks like reuse of an invalidated messages generator.

To see the issue in action, add a few print statements to the example code, run it and trigger a re-connect:

```
async def main():
    client = aiomqtt.Client("test.mosquitto.org")
    interval = 5  # Seconds
    while True:
        try:
            print(f"Entering client context...")
            async with client:
                    print(message.payload)
                print(f"Messages generator ends")
        except aiomqtt.MqttError:
            print(f"Connection lost; Reconnecting in {interval} seconds ...")
            await asyncio.sleep(interval)
```

This first time through, the async for loop works as expected, but when the client is re-used after a loss of connection, the messages generator ends and the code will loop connecting then immediately disconnecting.

Sorry, I'm not sure where the fix for this ought to go, nor what the proper use of the client.messages should to be - but the following change seems to work:

   - in client.__init__() set client.messages to None
   - in client.__aenter__() assign client.messages to client._messages()
   - in client.__aexit__() assign client.messages to None
 
Alternatively, adjusting the example code for loop as follows also works:

`async for message in client._messages():`

Setup:

   - Python 3.11.2 / Debian 12 amd64
   - paho-mqtt 1.6.1

spacemanspiff2007:
Fixes #268

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -716,9 +716,8 @@ class Client:
         tb: TracebackType | None,
     ) -> None:
         """Disconnect from the broker."""
-
         self.messages = None
-        
+
         if self._disconnected.done():
             # Return early if the client is already disconnected
             if self._lock.locked():
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -233,7 +233,7 @@ class Client:
         if max_queued_incoming_messages is None:
             max_queued_incoming_messages = 0
         self._queue = queue_type(maxsize=max_queued_incoming_messages)
-        self.messages = self._messages()
+        self.messages: AsyncGenerator[Message, None] | None = None
 
         # Semaphore to limit the number of concurrent outgoing calls
         self._outgoing_calls_sem: asyncio.Semaphore | None
@@ -705,6 +705,8 @@ class Client:
         # Reset `_disconnected` if it's already in completed state after connecting
         if self._disconnected.done():
             self._disconnected = asyncio.Future()
+
+        self.messages = self._messages()
         return self
 
     async def __aexit__(
@@ -714,6 +716,9 @@ class Client:
         tb: TracebackType | None,
     ) -> None:
         """Disconnect from the broker."""
+
+        self.messages = None
+        
         if self._disconnected.done():
             # Return early if the client is already disconnected
             if self._lock.locked():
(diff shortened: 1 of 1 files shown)


spacemanspiff2007:
I tried to add it but I am not fully sure if everything is in the right place ...

empicano:
Thanks for another PR from you! 🎸

> I'm not sure if allowing None for client.messages is a good idea.

I see this now as well. It also messes up the typing (I'm no pro with typing in Python, so could be that we can do this differently as well). The underlying problem is that generators in Python are invalidated once they throw an exception. It's possible to hack around this, but that will only be confusing in the end.

I think the way forward could be to rename `Client._messages()` to `Client.messages()` and use it like:

```python
    async for message in client.messages():
        print(message.payload)
```

Which is a breaking change. That's on me for not catching this issue when I implemented v2.0.0, sorry about that!

I think the best way for now is to use `Client._messages()`, what do you think? And I'll keep the breaking change in mind for v3.0.0.

spacemanspiff2007:
For me this is not an [issue](https://github.com/spacemanspiff2007/HABApp/blob/5ccb2d6f69e7917a766cac60ac25db5afa2919e6/src/HABApp/mqtt/connection/handler.py#L71) because I'm using a workaround.

However I think `Client.messages()` makes more sense than `Client.messages`.
The only graceful migration is to rename the `Client.messages` altogether.

I don't think using `Client._messages()` is a good idea and I think this fix definitely deserves a version bump and should be done asap. If properly documented in the release notes this should not be an issue.

vvanglro:
It seems that adding reusable features pits us.

This means that in the future, when initializing the Client class, all the class properties that need to be instantiated need to be taken care of to make sure that reusable does not cause problems.

spacemanspiff2007:
Imho ``.messages()`` makes more sense because it aligns e.g. with ``dict.keys()``.
If you prefer ``.messages`` then a proper type hint should be supplied. That's difficult because you can fall back to `None` which means every user has to check for `None` before accessing `.messages`. If in `__init__` only the hint is provided the client will raise an `AttributeError` when accessing `.messages` outside the context manager which is strange, too.

empicano:
I believe that this is fixed with #312. I only just now saw the additional commit with the migration suggestion you made, `@property` was a good idea, sorry about the oversight! Please let me know if there's anything left unsolved.

Thank you for the work you put into this PR nonetheless!

> Imho .messages() makes more sense because it aligns e.g. with dict.keys()

That's a really good example, `.messages()` also returns a dynamic view of the message queue so that fits well, I agree 👍

PR #273, opened 2024-02-02
fix: aiomqtt has no attribute __version__

### LINKED ISSUE #272: AttributeError: module 'aiomqtt' has no attribute '__version__'
```py
import aiomqtt
aiomqtt.__version__
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
AttributeError: module 'aiomqtt' has no attribute '__version__'
```

vvanglro:
Close #272 .

### DIFF
--- a/aiomqtt/__init__.py
+++ b/aiomqtt/__init__.py
@@ -10,6 +10,10 @@ from .exceptions import MqttCodeError, MqttError, MqttReentrantError
 from .message import Message
 from .topic import Topic, TopicLike, Wildcard, WildcardLike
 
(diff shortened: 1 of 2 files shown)


empicano:
LGTM 👍 I added the version requirement `>=1.0.0` for `poetry-dynamic-versioning` in `pyproject.toml` to make sure that the version tuple is also updated, because that didn't work for me at first. Thanks for another great PR, @vvanglro! 😊

PR #274, opened 2024-02-05
fix connection connection initialization handling

### LINKED ISSUE #269: Impossible to reuse client in case of connection timeout
I'm using almost the same code for reconnects as in docs:
```
async def main():
    client = aiomqtt.Client(broker, port, identifier = client_id, clean_session = True)
    reconnect_interval = 5
    while True:
        try:
            print(f"Connecting to {broker}:{port}")
            async with client:
                await client.subscribe(topic, qos = 1)
                print(f"Subscribed to {topic}")
                    pass
        except aiomqtt.MqttError as err:
            print(f"Connection lost: {err};")
            print(f"Reconnecting in {reconnect_interval} seconds ...")
            await asyncio.sleep(reconnect_interval)

asyncio.run(main())
```

If broker due to some problems cannot respond in given timeout aiomqtt gets `Operation timed out` error and client cannot be reused after that.

Example log:
```
Connecting to broker.local:1883
Connection lost: Operation timed out;
Reconnecting in 5 seconds ...
Connecting to broker.local:1883
Connection lost: The client context manager is reusable, but not reentrant;
Reconnecting in 5 seconds ...
Connecting to broker.local:1883
Connection lost: The client context manager is reusable, but not reentrant;
```

With other error like `Connection refused` if works fine:
```
Connecting to broker.local:1883
Connection lost: [Errno 111] Connection refused;
Reconnecting in 5 seconds ...
Connecting to broker.local:1883
Connection lost: [Errno 111] Connection refused;
Reconnecting in 5 seconds ...
```

I expect that regardless of connection error, client can be reused for reconnection

4s1:
Connection errors in __aenter__() which happen while _wait_for(selt._connected,...) were not catched, leaving a stale _lock behind, which made client non reusable.

Pull waiting for connected event into try/except block, so client is reusable in event of an error while trying to connect.

fixes #269

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -696,12 +696,12 @@ class Client:
                 self._properties,
             )
             _set_client_socket_defaults(self._client.socket(), self._socket_options)
+            await self._wait_for(self._connected, timeout=None)
         # Convert all possible paho-mqtt Client.connect exceptions to our MqttError
         # See: https://github.com/eclipse/paho.mqtt.python/blob/v1.5.0/src/paho/mqtt/client.py#L1770
-        except (OSError, mqtt.WebsocketConnectionError) as exc:
+        except (OSError, mqtt.WebsocketConnectionError, MqttConnectError) as exc:
             self._lock.release()
             raise MqttError(str(exc)) from None
-        await self._wait_for(self._connected, timeout=None)
         # Reset `_disconnected` if it's already in completed state after connecting
         if self._disconnected.done():
             self._disconnected = asyncio.Future()


empicano:
Hi there, thanks for working on this! 😎

As far as I understand, #269 mentioned two different exceptions. First, `MqttConnectError` thrown by the `on_connect` callback (which is handled in this PR) and second, `MqttError` thrown due to a timeout by `wait_for` (not yet handled).

Currently, the `MqttConnectError` would be cast to a `MqttError` which is a breaking change. I think it'd be better to handle and reraise both exceptions in their own try/except block, something like this (not tested):

```python
try:
    await self._wait_for(self._connected, timeout=None)
except MqttError, MqttConnectError:
    # Release lock and reraise any exception raised waiting on CONNACK
    self._lock.release()
    raise
```

It'd be great if you added test cases for these as well, so we/I don't accidently regress this again. To simulate the `MqttConnectError` thrown by `on_connect` you could use `client._connected.set_exception(MqttConnectError)`. There are a few tests that already do something similar with the `_disconnected` future.

empicano:
I'll close this, because I believe this was fixed with #285.

Thank you nonetheless for this PR and your great comments on the corresponding issue!

Issue #28, opened 2020-12-28
How to deal with LWT and DISCONNECT reason


sveinse:
I'm working out of the advanced example on the front page and was working on adding a last will testament (LWT). However, I noticed that `asyncio_mqtt.Client` with its usage of context manager is always cleaning up the connection by sending DISCONNECT. This has the effect that the LWT will not be sent out by the broker, when the program is handling any exceptions including keyboard interrupts. The only way to get LWT to work is to kill the program hard.

Consequently a "manual" LWT publish had to be added to ensure a LWT-like behavior on program exit. This was not immediately apparent, so it might be worth mentioning in the docs at some point.

I would propose that the `asyncio_mqtt.Client.__aexit__` had some awareness (setable flag in class?) of the exit status and corresponding `.disconnect()` had support for non-zero exit code in order for the broker to sent an ordinary LWT.

```python
# Create a LWT
will = Will("some/topic", payload="offline", qos=2, retain=False)

# Connect to the MQTT broker
client = Client("test.mosquitto.org", will=will)
await stack.enter_async_context(client)

# Implement manual LWT
stack.push_async_callback(client.publish, "some/topic", "offline")
```

frederikaalund:
Thanks the posting this issue. Let me have a look. :)

> However, I noticed that asyncio_mqtt.Client with its usage of context manager is always cleaning up the connection by sending DISCONNECT.

Yes, graceful disconnects are per design.

> This has the effect that the LWT will not be sent out by the broker, when the program is handling any exceptions including keyboard interrupts. The only way to get LWT to work is to kill the program hard.

This is indeed the intended *default* behaviour. Specifically, we consider a keyboard interrupt an intended disconnect. Hence the server shouldn't send the will message. I emphasize default, because I think it's a good idea to change the API to allow for custom behaviour.

## Short Note on the Will Feature

As far as I'm aware, the server should only send the will message on an unintended disconnect. Quoting the [MQTT Version 5.0 Specification, section 3.1.2.5 Will Flag](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901040):

> Situations in which the Will Message is published include, but are not limited to:
>  * An I/O error or network failure detected by the Server.
>  * The Client fails to communicate within the Keep Alive time.
>  * The Client closes the Network Connection without first sending a DISCONNECT packet with a Reason Code 0x00 (Normal disconnection).
[...]

PR #287, opened 2024-03-22
Implement automatic reconnection

empicano:
Hi there @frederikaalund @JonathanPlasse 😊

Frederik [drafted how reconnection could look like](https://github.com/sbtinstruments/aiomqtt/discussions/240#discussioncomment-6767115) a while ago already (thank you again, master of asyncio 🙏😄). I finally had some time to hack around with this! Some thoughts:

- I'd like to avoid having two clients (one that reconnects and one that doesn't) because I suspect that this is going to be confusing to use and lead to more maintenance in the future. I'm thus trying to implement it as a `reconnect=True` parameter to the client and leave all method signatures the same.
- Consequently, this means that all publish/subscribe/etc. calls have to be asynchronous, contrary to Frederik's original design. Instead of queuing the calls up, I'm thinking of blocking them and returning only when the connection returns and the call goes through.
- This would eliminate the flip side that you mentioned as well, Frederik: "The user does not know if/when their events actually go through."

For now, I've implemented the reconnection background task and adapted the `publish` method to wait until the connection returns and the message could actually be published (both are probably still full of bugs). You can play around with it by shutting a local MQTT broker on and off (e.g. with `./scripts/develop`) and by running:

```python
import aiomqtt
import asyncio


async def main():
    client = aiomqtt.Client("localhost", 1883, reconnect=True, timeout=2)
    async with client:
        while True:
            await asyncio.sleep(1)
            print("Publishing message...")
            await client.publish("test", "Hello, World!")
            print("Message published.")


asyncio.run(main())
```

I'm still thinking about how to deal with existing subscriptions and last wills. We probably have to resend them when `clean_session=True`, otherwise they will cease to exist without notice to the user after a reconnection in the background.

Happy to hear what your thoughts are on this (or anyone else's who wants to chip in) 😊

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -129,6 +129,8 @@ class Client:
         password: The password to authenticate with.
         logger: Custom logger instance.
         identifier: The client identifier. Generated automatically if ``None``.
+        reconnect: If ``True``, the client will automatically reconnect to the broker
+            if the connection is lost. Defaults to ``False``.
         queue_type: The class to use for the queue. The default is
             ``asyncio.Queue``, which stores messages in FIFO order. For LIFO order,
             you can use ``asyncio.LifoQueue``; For priority order you can subclass
@@ -181,6 +183,7 @@ class Client:
         password: str | None = None,
         logger: logging.Logger | None = None,
         identifier: str | None = None,
+        reconnect: bool = False,
         queue_type: type[asyncio.Queue[Message]] | None = None,
         protocol: ProtocolVersion | None = None,
         will: Will | None = None,
@@ -206,6 +209,7 @@ class Client:
     ) -> None:
         self._hostname = hostname
         self._port = port
+        self._reconnect = reconnect
         self._keepalive = keepalive
         self._bind_address = bind_address
         self._bind_port = bind_port
@@ -225,7 +229,10 @@ class Client:
         self._pending_unsubscribes: dict[int, asyncio.Event] = {}
         self._pending_publishes: dict[int, asyncio.Event] = {}
         self.pending_calls_threshold: int = 10
+
+        # Background tasks
         self._misc_task: asyncio.Task[None] | None = None
+        self._reconnection_task: asyncio.Task[None] | None = None
 
         # Queue that holds incoming messages
         if queue_type is None:
@@ -432,9 +439,17 @@ class Client:
             **kwargs: Additional keyword arguments to pass to paho-mqtt's publish
                 method.
         """
-        info = self._client.publish(
-            topic, payload, qos, retain, properties, *args, **kwargs
-        )  # [2]
+        while True:
+            info = self._client.publish(
+                topic, payload, qos, retain, properties, *args, **kwargs
+            )  # [2]
+            if not (info.rc == mqtt.MQTT_ERR_NO_CONN and self._reconnect):
+                break
+            while True:
+                with contextlib.suppress(asyncio.CancelledError):
+                    await self._connected
+                    break
+                self._connected = asyncio.Future()
         # Early out on error
         if info.rc != mqtt.MQTT_ERR_SUCCESS:
             raise MqttCodeError(info.rc, "Could not publish message")
@@ -677,43 +692,65 @@ class Client:
         while self._client.loop_misc() == mqtt.MQTT_ERR_SUCCESS:
             await asyncio.sleep(1)
 
+    async def _connect(self) -> None:
+        """Connect to the broker. Retry indefinitely if self._reconnect is True."""
+        while True:
+            try:
+                try:
+                    loop = asyncio.get_running_loop()
+                    # [3] Run connect() within an executor thread, since it blocks on socket
+                    # connection for up to `keepalive` seconds: https://git.io/Jt5Yc
+                    await loop.run_in_executor(
+                        None,
+                        self._client.connect,
+                        self._hostname,
+                        self._port,
+                        self._keepalive,
+                        self._bind_address,
+                        self._bind_port,
+                        self._clean_start,
+                        self._properties,
+                    )
+                    _set_client_socket_defaults(self._client.socket(), self._socket_options)
+                # Convert all possible paho-mqtt Client.connect exceptions to our MqttError
+                # See: https://github.com/eclipse/paho.mqtt.python/blob/v1.5.0/src/paho/mqtt/client.py#L1770
+                except (OSError, mqtt.WebsocketConnectionError) as exc:
+                    raise MqttError(str(exc)) from None
+                await self._wait_for(self._connected, timeout=None)
+                self._logger.info("Successfully connected to the broker.")
+                break
+            except MqttError:
+                # Reset internal state if the connection attempt failed
+                if self._connected.done():
+                    self._connected = asyncio.Future()
+                if self._disconnected.done():
+                    self._disconnected = asyncio.Future()
+                if not self._reconnect:
+                    self._lock.release()
+                    raise
+                self._logger.warning("Failed to connect. Trying again in 2 seconds...")
+                await asyncio.sleep(2)
+
+    async def _reconnection(self) -> None:
+        """Reconnect when the connection is lost."""
+        while True:
+            with contextlib.suppress(MqttError):
+                await self._disconnected
+            self._logger.warning("Connection lost. Reconnecting...")
+            self._connected = asyncio.Future()
+            self._disconnected = asyncio.Future()
+            await self._connect()
+
     async def __aenter__(self) -> Self:
         """Connect to the broker."""
         if self._lock.locked():
             msg = "The client context manager is reusable, but not reentrant"
             raise MqttReentrantError(msg)
         await self._lock.acquire()
-        try:
-            loop = asyncio.get_running_loop()
-            # [3] Run connect() within an executor thread, since it blocks on socket
-            # connection for up to `keepalive` seconds: https://git.io/Jt5Yc
-            await loop.run_in_executor(
-                None,
-                self._client.connect,
-                self._hostname,
-                self._port,
-                self._keepalive,
-                self._bind_address,
-                self._bind_port,
-                self._clean_start,
-                self._properties,
-            )
-            _set_client_socket_defaults(self._client.socket(), self._socket_options)
-        # Convert all possible paho-mqtt Client.connect exceptions to our MqttError
-        # See: https://github.com/eclipse/paho.mqtt.python/blob/v1.5.0/src/paho/mqtt/client.py#L1770
-        except (OSError, mqtt.WebsocketConnectionError) as exc:
-            self._lock.release()
-            raise MqttError(str(exc)) from None
-        try:
-            await self._wait_for(self._connected, timeout=None)
-        except MqttError:
-            # Reset state if connection attempt times out or CONNACK returns negative
-            self._lock.release()
-            self._connected = asyncio.Future()
-            raise
-        # Reset `_disconnected` if it's already in completed state after connecting
-        if self._disconnected.done():
-            self._disconnected = asyncio.Future()
+        await self._connect()
+        # Start the reconnection task
+        if self._reconnect:
+            self._reconnection_task = asyncio.create_task(self._reconnection())
         return self
 
     async def __aexit__(
@@ -723,8 +760,10 @@ class Client:
         tb: TracebackType | None,
     ) -> None:
         """Disconnect from the broker."""
+        if self._reconnect:
+            self._reconnection_task.cancel()
+        # Return early if the client is already disconnected
         if self._disconnected.done():
-            # Return early if the client is already disconnected
             if self._lock.locked():
                 self._lock.release()
             if (exc := self._disconnected.exception()) is not None:


frederikaalund:
Thank you for looking into this! 👍 Automatic reconnect is definitely on the wish list of many of our users. :)

I do like the idea to have "everything in one client" as opposed to my approach with a separate low-level and high-level client. At least the single-client approach is easier to grok for our users. 👍 

I looked through the code, and it looks good. I'm (as always) a little bit concerned about adding extra internal state to our client (the new reconnect task). Increases the maintenance burden. In any case, for this feature I don't think we can avoid it *if* we want to use a single client. 😄

*Aside:* There is a lot of "reset state" going on (re-creating the futures). That's not your fault (or the fault of this PR) but my own fault (using futures in the first place). 😅 I strongly suggest that we look at alternatives for all the internal futures and background tasks. I'm thinking anyio (to no ones surprise I guess). :)

> I'm still thinking about how to deal with existing subscriptions and last wills. We probably have to resend them when `clean_session=True`, otherwise they will cease to exist without notice to the user after a reconnection in the background.

That is a concern of mine as well! Indeed, we would have to save all the subscriptions inside the client and then "resubscribe" when the connection is back online. These subscriptions are even more state to manage. :)

---

Here is a suggestion that is a slight variation of the current approach:

 * Single *public* client (that now can also automatically reconnect)
   - Rename the existing `Client` to `_InternalClient` (or something like that)
   - Create a new `Client` class *with the exact same API as the existing client* (our users won't see a difference!)
   - Use [streams and broadcast mechanisms in the new `Client`](https://github.com/sbtinstruments/aiomqtt/discussions/240#discussioncomment-6767115) to implement the reconnect logic.
 
This way, we get the separation of high-level and low-level and the easier maintenance that follows. It also allows allows to replace the `_InternalClient` with a new implementation in the future (maybe written in rust, using anyio, or something else).

> Consequently, this means that all publish/subscribe/etc. calls have to be asynchronous, contrary to Frederik's original design. Instead of queuing the calls up, I'm thinking of blocking them and returning only when the connection returns and the call goes through.

I see the benefits of this approach! 👍 I suggest that we do both 😄  That is, to have both `async def publish()` and `def publish_nowait()`. This mimicks how both asyncio and anyio handles the "do you want to wait or not" question. In turn, it let's the user get to decide whether they want to wait for the connection or not. :)


Again, thank you for looking into this issue and very well done on the draft implementation. 👍  Let me know what you think of my comments above and do say if you have any questions. 😄

empicano:
Thanks for your thoughts on this! Your reviews are one of the main reasons this project is so fun for me 😉

You're right about the internal state getting slightly out of control. While implementing this draft I already got bitten resetting futures while they were awaited elsewhere (which raises `CancelledError`) 😅

> Use streams and broadcast mechanisms in the new `Client` to implement the reconnect logic.

I'm not sure I understand this point. Do you mean using streams and broadcast for the `_nowait` variants like you did in your example, or actually a different way to implement the reconnection? Apart from that, I like the idea of bringing more structure into it with the low-level client class 🙂

I'm on board with the `publish` and `publish_nowait` design as well 👍 I can imagine how these can be implemented in the reconnection case, but I wonder how `publish_nowait` could work when the client shouldn't reconnect. (How) Do we fail in the case the client is disconnected? Given that we probably work through the queued messages in the background without awaiting the task anywhere, throwing an exception won't show to the user, or can we propagate it somehow?

empicano:
Thanks for elaborating, I think I understand what you mean now 😊 I'll play around with streams and broadcasts to understand them a little better and see how much work the second option would bring! You convinced me that that's the better option 😄

I'll report back once I have more, could take a bit, though, as I'm busy the next few days 😋

spacemanspiff2007:
I just stumbled across this so sorry if I am late to the party. From my experience it might be better to split the client into a low level and high level because as mentioned it makes maintenance much easier, however I have no strong opinion on that.

While looking through the PR I am not sure I understood everything correctly:
It seems like publish will never return until the client is connected again possibly locking up the program. This is imho very unexpected. 
[Here](https://github.com/spacemanspiff2007/sml2mqtt/blob/78d9c070e1bbeb140e504d03cbfc9ab68235a0d6/src/sml2mqtt/mqtt/mqtt.py#L92) I have a small program that only publishes and I just discard all messages on disconnect. I even made publish non-blocking and even non async through the use of a `Queue`.
On the other hand `_messages` still raises an error when the connection is lost. I think it would be good if the behavior of these two methods is aligned.
In case of a disconnect there will be a retry every two seconds. It would be nice if there is the possibility to provide something (e.g. generator) that backs up gracefully (e.g. 2, 4, 8 ... 300) secs delay.
Additionally it would be nice to have the possibility to specify an additional message that gets sent on graceful disconnect on the last will topic because last will is only used on abnormal disconnect.

I think he usage of the client is hard because there are multiple places where the disconnect error can occur. Typically I have a `publish` and a `messages` worker task and both can be the cause for a reconnect which I have to sync back to the task where I create the client and do the connect. Do you know by chance a good solution for that? I've come up with a quite complex solution, a stateful connection manager which connects and then creates the tasks. If one task throws an `MqttError` I cancel both tasks and try to trigger a reconnect. 
Example from one application: 
[connection handler](https://github.com/spacemanspiff2007/HABApp/blob/5ccb2d6f69e7917a766cac60ac25db5afa2919e6/src/HABApp/mqtt/connection/handler.py#L68)
[publish task](https://github.com/spacemanspiff2007/HABApp/blob/5ccb2d6f69e7917a766cac60ac25db5afa2919e6/src/HABApp/mqtt/connection/publish.py#L34)
[messages task](https://github.com/spacemanspiff2007/HABApp/blob/5ccb2d6f69e7917a766cac60ac25db5afa2919e6/src/HABApp/mqtt/connection/subscribe.py#L154)

empicano:
Thanks for chipping in and sorry for the late reply! 🙂

> It seems like publish will never return until the client is connected again possibly locking up the program. This is imho very unexpected. [...] On the other hand _messages still raises an error when the connection is lost. I think it would be good if the behavior of these two methods is aligned.

I agree that `publish` and `_messages` should behave the same. If I understand you right, you're suggesting that the client throws an exception in both cases until the client has successfully reconnected in the background? Of the top of my head, I can't think of a way to know when to retry calling `publish` or `_messages` in that case without accessing internals (which I'd like to avoid). Waiting a fixed amount of time seems crude, and with the other option we could always use [`asyncio.timeout`](https://docs.python.org/3/library/asyncio-task.html#timeouts) if we want to set an upper limit for how long we want to wait.

What benefits do you see from failing right away?

> In case of a disconnect there will be a retry every two seconds. It would be nice if there is the possibility to provide something (e.g. generator) that backs up gracefully (e.g. 2, 4, 8 ... 300) secs delay.

Good idea 👍

> Additionally it would be nice to have the possibility to specify an additional message that gets sent on graceful disconnect on the last will topic because last will is only used on abnormal disconnect.

I noticed that, too. Currently we can only set the last will on client initialization. Maybe we can add a function to set the last will dynamically in the future.

spacemanspiff2007:
> I agree that `publish` and `_messages` should behave the same. If I understand you right, you're suggesting that the client throws an exception in both cases until the client has successfully reconnected in the background? Of the top of my head, I can't think of a way to know when to retry calling `publish` or `_messages` in that case without accessing internals (which I'd like to avoid). Waiting a fixed amount of time seems crude, and with the other option we could always use [`asyncio.timeout`](https://docs.python.org/3/library/asyncio-task.html#timeouts) if we want to set an upper limit for how long we want to wait.

Maybe expose some client status through a property - e.g. ``client.is_connected``?

I am suggesting that [publish](https://github.com/sbtinstruments/aiomqtt/pull/287/files#diff-55d3e0448cfc11b93eaa5c07050506efe4eaf2fcb108f06751e9dd106bfbac1bR442) should not block when the client is disconnected because it's very unexpected. If I have a small program likes this
````python
while True:
    await read_sensors()
    try:
        await client.publish(...)
    except Exception:
        pass
    await write_sensor_to_local_db()
````
it will stop reading sensors when the client is disconnected and it's not clear at all that it will behave like that.

---

I think it depends on the program how publish should behave and currently I can think of three different desired behaviors
- Block until published
- Notify the calling code that the client is disconnected e.g. through return value or exception and discard the message
- Return some object which then can be 
  - awaited - which will wait until the message is published
  - canceled - which will not try to publish it any more
  - queried - which will return the current status (published, pending), maybe the connection status, etc.

The third one is the most generic one because with it it's easy to implement the first two behaviors and I would expect a high level client to return something like that.



> Maybe we can add a function to set the last will dynamically in the future.

I currently don't understand how a function will provide a message when gracefully disconnecting.
[Here](https://github.com/spacemanspiff2007/sml2mqtt/blob/78d9c070e1bbeb140e504d03cbfc9ab68235a0d6/src/sml2mqtt/mqtt/mqtt.py#L108) I have a dedicated publish task and when I cancel the task (graceful disconnect) I catch that and publish the message.

empicano:
> Maybe expose some client status through a property - e.g. client.is_connected?

I'd like to avoid that, if possible. Exposing internals like that often leads to weird problems down the line.

> I am suggesting that publish should not block when the client is disconnected because it's very unexpected.

I think I see why now. When `reconnect=False`, `publish` blocks until the message is published **or fails when the connection is lost**. With the initial idea and `reconnect=True`, `publish` blocks until the message is published, but doesn't fail when the connection is lost. The most intuitive design would be that `publish` behaves exactly the same in both cases, meaning also in the case of disconnection, right?

If we think about how an asynchronous `publish` and a synchronous `publish_nowait` with Frederik's suggestion could work in the two `reconnect=True/False` cases, with the above the most logical would thus be:

1. `reconnect=True/False` & `publish`: block until the message is sent, fail on disconnection
1. `reconnect=True/False` & `publish_nowait`: queue the message and return immediately, no failure on disconnection

However, `publish_nowait` doesn't make too much sense in case of `reconnect=False` because if that's all we use, we'll append and append messages to the queue without ever noticing that the client has disconnected. And if we don't block `publish` until reconnection in the `reconnect=True` case, `publish_nowait` doesn't really have a benefit any more, either.

I think you convinced me that a better design would be to implement only an asynchronous `publish`, but also make it fail on disconnection in the `reconnect=True` case. What I miss is a good way to know when to retry a failed publish in the case of `reconnect=True`.

> I think it depends on the program how publish should behave and currently I can think of three different desired behaviors

Returning something wouldn't work very well in the case of the `_messages` generator, so I'd focus on the second point and think about how we can achieve the first one on top.

One option would be to do the reconnect manually, so we could do `await client.reconnect()` to achieve something similar to `client.is_connected`. That would eliminate the whole discussion about `publish` in the `reconnect=False` and `reconnect=True` cases, because there won't be a `reconnect` client argument anymore. We're basically pulling the reconnection from the background into a manual function call. This seems pretty flexible, I think I like this.

This turned out to be a long rambling, I hope it's understandable 😄 Do say if something's not clear, and let me know what you think! Again, thanks for chipping in, it helps quite a lot to have other perspectives on this. You make some good points, not only here, but also in the other issues and PRs that you're involved with 👍

---

> Here I have a dedicated publish task and when I cancel the task (graceful disconnect) I catch that and publish the message.

Ah, I think I understand what you mean now 👍 #28 is related, and a very interesting discussion. There already has been a solution proposed, I'm interested to hear what you think about it. Maybe you are up to do a PR? 🙂

spacemanspiff2007:
> The most intuitive design would be that `publish` behaves exactly the same in both cases, meaning also in the case of disconnection, right?

This is not very nice but not my main issue. With `reconnect=True` the application will be stuck in `await client.publish()` **until** the client is connected again. In my example above the program also writes the sensor values to a database. If the broker goes down all values will be lost until the broker is up again and there will be a gap in the db values.
This behavior is very unexpected and it's not clear that this will happen from looking at the code snippet.

> However, `publish_nowait` doesn't make too much sense in case of `reconnect=False` because if that's all we use, we'll append and append messages to the queue without ever noticing that the client has disconnected.

I would have expected `publish_nowait` to discard messages when the client is not connected. On disconnect the queue should be emptied.
In the sml2mqtt program I implemented it an same manner:
[publish](https://github.com/spacemanspiff2007/sml2mqtt/blob/78d9c070e1bbeb140e504d03cbfc9ab68235a0d6/src/sml2mqtt/mqtt/mqtt.py#L129) only works when the client is connected and on disconnect the queue is [discarded](https://github.com/spacemanspiff2007/sml2mqtt/blob/78d9c070e1bbeb140e504d03cbfc9ab68235a0d6/src/sml2mqtt/mqtt/mqtt.py#L125).
A behavior like this would make small programs that continuously collect data very easy to implement.



> Returning something (third bullet point) wouldn't work very well in the case of the `_messages` generator, so I'd focus on the second point and think about how we can achieve the first one on top.

Could you please elaborate why this would be an issue? Currently I fail to see how the `publish` call interferes with `messages` which is used to process incoming messages.
I still think from a usability perspective the is the best option because it covers all use cases and makes it very clear from the calling code what is happening:

````python
# create obj
msg = client.publish(...)

msg.cancel_publish()

# status or flags
msg.status
msg.is_published

# wait for publish
await msg.published()
await msg.published(reconnect=True)  # Could be on a per message basis
````



> One option would be to do the reconnect manually, so we could do `await client.reconnect()` to achieve something similar to `client.is_connected`. That would eliminate the whole discussion about `publish` in the `reconnect=False` and `reconnect=True` cases, because there won't be a `reconnect` client argument anymore. We're basically pulling the reconnection from the background into a manual function call. This seems pretty flexible.

What would your goal be with that and can you make a pseudo example? Do I as a caller still have to catch an error and manually call reconnect? Do I have to do it both in `publish` and `messages`?



> Again, thanks for chipping in, it helps quite a lot to have other perspectives on this. You make some good points, not only here, but also in the other issues and PRs that you're involved with 👍

Thanks - that means a lot!

carstenandrich:
@empicano Opened discussion #316 regarding a possibly slightly simpler implementation approach.

mirko:
I'm coming from https://github.com/empicano/aiomqtt/issues/334#issuecomment-2341342920 - and would just like to provide some input from my perspective, as I was asked to. Keep in mind, though, that I'm having a more low-level background (microcontrollers, C) and I'm only having one - my above referenced - use-case in mind.

Assuming the `connect()`/`disconnect()` methods are not coming back, I - as a dev using this lib - would expect publish calls to fail and not (a) being queued and potentially fail later and (b) wait (potentially indefinitely) until completed.

That being said, I`d expect if `await client.publish()` finishes, the message was sent (or at least written on the underlying socket). And I'd expect (different) exceptions if it wasn't published (in time) - to individually be able to react to network issues and potentially (indefinitely) try again.

Network issues while subscribing() appears a more challenging topic at first glance to me. Depending on the use-case I might definitely wanna know if I potentially missed a message - or I don't care and expect the loop iterating over messages, most likely running in its dedicated task, to keep going at best effort. What really doesn't help is how exceptions thrown in asyncio tasks need to be handled, which - according to my knowledge - only works with a global asyncio exception handler. I guess it's still better than just leaving the user in the dark whether messages were potentially unhandled due to network issues, though.

PR #302, opened 2024-05-22
[WIP] transition to anyio

smurfix:
As per https://github.com/sbtinstruments/aiomqtt/discussions/226 here's a draft of an anyio-ified aiomqtt. It abuses paho-mqtt 2.0 by way of subclassing its `Client` class.

No more threads. No more issues with locking. Trio compatibility.

Tests pass (one failure on asyncio) but there's still a lot to do; checking robustness and handling reconnections and getting typing back up to standard and finishing the paho callback API v2 transition and support for paho-mqtt 2.1 and general clean-up and support for more MQTTv5 features and whatnot.

I need this code in order to replace my `moat-mqtt` package  (which is a hbmqtt port to anyio that shows its age). No guarantees but I plan to keep poking at this.

### DIFF
--- a/aiomqtt/queue.py
+++ b/aiomqtt/queue.py
@@ -1,0 +1,88 @@
+"""
+trio/anyio no longer have queues, but sometimes a memory object stream is
+unwieldy. Thus this re-implemens a simple queue using
+`anyio.create_memory_object_stream`.
+"""
+
+from __future__ import annotations
+
+import anyio
+from anyio import create_memory_object_stream
+
+from outcome import Error, Value
+
+from typing import TYPE_CHECKING  # isort:skip
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable
+
+import logging  # isort:skip
+
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "Queue",
+]
+
+
+class Queue:
+    """
+    Queues have been replaced in trio/anyio by memory object streams, but
+    those are more complicated to use.
+
+    This Queue class simply re-implements queues on top of memory object streams.
+    """
+
+    def __init__(self, length=1):
+        self._s, self._r = create_memory_object_stream(length)
+
+    async def put(self, x):
+        """Send a value, blocking"""
+        await self._s.send(Value(x))
+
+    def put_nowait(self, x):
+        """Send a value, nonblocking"""
+        self._s.send_nowait(Value(x))
+
+    async def put_error(self, x):
+        """Send an error value, blocking"""
+        await self._s.send(Error(x))
+
+    async def get(self):
+        """Get the next value, blocking.
+        May raise an exception if one was sent."""
+        res = await self._r.receive()
+        return res.unwrap()
+
+    def get_nowait(self):
+        """Get the next value, nonblocking.
+        May raise an exception if one was sent."""
+        res = self._r.receive_nowait()
+        return res.unwrap()
+
+    def qsize(self):
+        """Return the number of elements in the queue"""
+        return self._s.statistics().current_buffer_used
+
+    def empty(self):
+        """Check whether the queue is empty"""
+        return self._s.statistics().current_buffer_used == 0
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        res = await self._r.__anext__()  # pylint: disable=E1101
+        return res.unwrap()
+
+    def close_sender(self):
+        """No more messages will be received"""
+        self._s.close()
+
+    close_writer = close_sender
+
+    def close_receiver(self):
+        """No more messages may be sent"""
+        self._r.close()
+
+    close_reader = close_receiver
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,10 +6,7 @@ from typing import Any
 import pytest
 
 
+# No longer needed, this also works with Trio
 @pytest.fixture
 def anyio_backend() -> tuple[str, dict[str, Any]]:
-    if sys.platform == "win32":
-        from asyncio.windows_events import WindowsSelectorEventLoopPolicy
-
-        return ("asyncio", {"policy": WindowsSelectorEventLoopPolicy()})
-    return ("asyncio", {})
+    return ("trio", {})
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ aiomqtt/_version.py
 *.DS_Store
 docs/_build
 reports
+
+# Debian packaging
+/debian/
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<h1 align="center">The idiomatic asyncio MQTT client 🙌</h1>
+<h1 align="center">The idiomatic async MQTT client 🙌</h1>
 <p align="center"><em>(formerly known as asyncio-mqtt)</em></p>
 <p align="center">
     <a href="https://github.com/sbtinstruments/aiomqtt/blob/main/LICENSE"><img alt="License: BSD-3-Clause" src="https://img.shields.io/github/license/sbtinstruments/aiomqtt"></a>
@@ -33,7 +33,7 @@ async with Client("test.mosquitto.org") as client:
         print(message.payload)
 ```
 
-aiomqtt combines the stability of the time-proven [paho-mqtt](https://github.com/eclipse/paho.mqtt.python) library with an idiomatic asyncio interface:
+aiomqtt combines the stability of the time-proven [paho-mqtt](https://github.com/eclipse/paho.mqtt.python) library with an idiomatic async interface:
 
 - No more callbacks! 👍
 - No more return codes (welcome to the `MqttError`)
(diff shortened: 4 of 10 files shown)


tests/conftest.py
```diff
@@ -6,10 +6,7 @@
...
 
+# No longer needed, this also works with Trio
 @pytest.fixture
 def anyio_backend() -> tuple[str, dict[str, Any]]:
-    if sys.platform == "win32":
-        from asyncio.windows_events import WindowsSelectorEventLoopPolicy
-
-        return ("asyncio", {"policy": WindowsSelectorEventLoopPolicy()})
-    return ("asyncio", {})
+    return ("trio", {})
```
agronholm, line 12: This means that the test suite will _only_ run against Trio now. The proper fix is to delete the fixture completely.

smurfix: That's temporary until the next anyio release and/or I can figure out the reason for the test failure.

smurfix: Removed.

smurfix:
Now includes a `subscription` method that lets you do "local" message processing:

```
    async with client.subscription("foo/bar") as msgs:
        async for m in msgs:
            await process(m)
```

empicano:
Hi there, very cool to see so much work on this! 👍 

This seems to be the go-ahead for our yearly discussion on this subject 😄 For reference, we had a PR in the past (#152) to switch to anyio that we ultimately decided to abandon. In discussion #44, people were reluctant about a switch to anyio, especially if the changes would impact the interface.

All that being said, we focus a lot on keeping aiomqtt as small and maintainable as possible. I can see an internal switch to anyio if it works towards that goal. In the best case, this would be non-breaking. The trio support would be a nice extra, but wouldn't make the switch worth it to me if it increases maintenance cost. I am also still on the edge if it's worth adding another dependency if we get e.g. task groups "for free" once we drop Python <3.11.

@frederikaalund, what are your thoughts on this? 🙂

---

> Now includes a subscription method that lets you do "local" message processing

We had something similar to this in the past with `unfiltered_messages` and `filtered_messages`. People got confused about how it works, among others the fact that this creates multiple queues, and we thus decided to change it to what we have now.

aiomqtt/queue.py
```diff
@@ -0,0 +1,88 @@
...
+unwieldy. Thus this re-implemens a simple queue using
+`anyio.create_memory_object_stream`.
+"""
+
+from __future__ import annotations
+
+import anyio
+from anyio import create_memory_object_stream
+
+from outcome import Error, Value
```
agronholm, line 12: I didn't see `outcome` in the dependencies. I also don't think it's worth the trouble to implement a wrapper on top of memory object streams.

agronholm:
Is it absolutely necessary to depend on the `paho-mqtt` client? Can't we instead create a sans-io implementation of the MQTT v5 / v3.1 protocols?

smurfix:
@agronholm Paho-MQTT at least has a working MQTTv5 implementation and complete data classes for all the fiddly bits; reimplementing all that is not exactly trivial, thus the straightest way to working threads-free async-only code was to anyio-ify it by way of a subclass. I'm the first person to admit that this subclass is way more intrusive than any sane programmer should be comfortable with, but you got to start somewhere.

I do plan to build a reasonable sans-io MQTT core eventually. Ideally I'd then be able to teach the current paho.mqtt.Client class to use that (should be a net reduction in code size …). Assuming that its maintainers are amenable to this, which I'm a bit skeptical about, but we'll see. In the meantime I've submitted a somewhat less intrusive clean-up patch to them.

smurfix:
@empicano asyncio taskgroups may have the same name, but not the same features of anyio taskgroups. For instance, they don't have cancel scopes; you need to cancel the task the group runs in, which is not the same thing. There are other differences which @agronholm is way more qualified than I am to educate people about. ;-)

As to changing the interface, my patch manages not to do that. Admittedly it does this by way of somewhat-invasively subclassing the Paho client but frankly the Paho code could do with a bit of improvement. :-/

As far as "filtered_messages", IMHO it's far more confusing, or rather detrimental to modularity, to require a central "async with client.messages()" loop to do the message dispatching. You can't do a "this task deals with temperatures and that task deals with humidity" pattern that way. It also requires more CPU.

IMHO it's a question of designing the right interface. The old `async with filtered_messages()` pattern required the caller to do the subscribe/loop-/unsubscribe dance instead of having the context manager handle it, which I can't see a good reason for. It also risked losing messages, because the filter was attached *after* subscribing instead of beforehand.

NB 9c6a7f4 documents the new `async with client.subscription()` interface. Personally I'd much rather use my example code than the equivalent (but longer and less modular, esp. when you add queue-full-error handling) example above it.

agronholm:
> @agronholm Paho-MQTT at least has a working MQTTv5 implementation and complete data classes for all the fiddly bits; reimplementing all that is not exactly trivial, thus the straightest way to working threads-free async-only code was to anyio-ify it by way of a subclass. I'm the first person to admit that this subclass is way more intrusive than any sane programmer should be comfortable with, but you got to start somewhere.
> 
> I do plan to build a reasonable sans-io MQTT core eventually. Ideally I'd then be able to teach the current paho.mqtt.Client class to use that (should be a net reduction in code size …). Assuming that its maintainers are amenable to this, which I'm a bit skeptical about, but we'll see. In the meantime I've submitted a somewhat less intrusive clean-up patch to them.
> 
> [eclipse/paho.mqtt.python#845](https://github.com/eclipse/paho.mqtt.python/pull/845)

Fair enough.

empicano:
I hope Frederik will still comment here, he can frankly review this better than me. I'll say that I'd be pretty neat to get rid of the threads and clean up the code, I'm just concerned about maintainability when adding as much code as we're already stretched thin as is. I agree that that sans-io "backend" would be really cool to see in aiomqtt 👍

---

On the interface topic:

> You can't do a "this task deals with temperatures and that task deals with humidity" pattern that way.

Maybe I misunderstand you here, I'd say [this example in the documentation](https://sbtinstruments.github.io/aiomqtt/subscribing-to-a-topic.html#multiple-queues) does that, or not? The flexibility of this approach was in fact one of the main reasons I pushed for the change.

Generally, I want this library to serve the most basic use case (which I think is a single task and queue) as well as possible, with flexibility to use it for as much as possible after that. I'd like to avoid having multiple queues by default, because when we still had `filtered_messages` we heared multiple times in issues and discussions that this led to confusion and problems.

Apart from that, I'd expect a fixed number of tasks working a single priority queue to fit better in most use cases where we need to process message concurrently, e.g. because the number of messages coming over different topics is often skewed. What do you think about that?

> IMHO it's a question of designing the right interface. The old async with filtered_messages() pattern required the caller to do the subscribe/loop-/unsubscribe dance instead of having the context manager handle it, which I can't see a good reason for. It also risked losing messages, because the filter was attached after subscribing instead of beforehand.

In my opinion, subscriptions and routing should be separate. I admit that context managers for subscriptions are elegant, but most of the time we either don't need to unsubscribe (clean session) or don't want to unsubscribe (persistent session) before disconnection. I want this library to support as many use cases as possible, and dynamic un/subscriptions are part of that. We can still have those via task cancellation with the proposed changes, but I'd like to find a simpler solution.

Let me propose an alternative idea to improve the message handling interface:

```python
router = aiomqtt.Router()


@router.match("humidity/+/inside/#")
async def handle(message, client, *args):  # automatically extract wildcards into *args
    print(message.payload)


async with aiomqtt.Client("test.mosquitto.org", routers=[router]) as client:
        await client.route(message)
```

Where we can process messages concurrently e.g. like this:


```python
router = aiomqtt.Router()


@router.match("humidity/+/inside/#")
async def handle(message, client, *args):  # automatically extract wildcards into *args
    print(message.payload)


async def work(client):
    async for message in client._messages():
        await client.route(message)


async with aiomqtt.Client("test.mosquitto.org", routers=[router]) as client:
    async with asyncio.TaskGroup() as tg:
        tg.create_task(work(client))
        tg.create_task(work(client))
```

(`.messages` is currently broken, for now we use `._messages()` to be able to use it multiple times.)

I opened #304 with a proof of concept. I'd be glad to hear what you think!

smurfix:
Well, the downside of separating subscription and routing is, as your example demonstrates, that the router needs to match every message to all your wildcard patterns. My code uses the fact that the server already did that work.

As to sans-io – in an ideal world that'd be part of Paho and thus not add to aiomqtt's maintainance burden. We'll see.

agronholm:
> Well, the downside of separating subscription and routing is, as your example demonstrates, that the router needs to match every message to all your wildcard patterns. My code uses the fact that the server already did that work.
> 
> As to sans-io – in an ideal world that'd be part of Paho and thus not add to aiomqtt's maintainance burden. We'll see.

Did I ever mention https://github.com/agronholm/mqttproto ?

smurfix:
> Did I ever mention [mqttproto](https://github.com/agronholm/mqttproto) ?

Umm no you didn't.

I'm afraid that this is as good a place as any to cut my losses and switch to that. If nothing else, its overhead is *way* smaller than anything we could reasonably massage the Paho code into.

The fact that the Paho people are very reluctant to accept patches doesn't help either.

PR #304, opened 2024-06-02
Improve message routing

empicano:
This proposes a simpler way to filter messages and structure message handling.

Unsorted points that I considered:

- This lets us split message handling into multiple files via multiple routers
- This lets us use the client inside a handler function, to e.g. publish a message back in a request/response fashion
- We can dynamically subscribe and unsubscribe
- The values of wildcards (`+`/`#`) of the topic filter are automatically available as `*args` in the handler function
- We still only have a single message queue (easier for newcomers, concurrency could be implemented as shown below, optionally with priority queue)
- We can still pass a non-default queue to the client to prioritize the handling of certain messages
- We still have flexibility to not use the routers, but handle the messages directly. Routers are a natural development, once the application gets too complex we can iteratively add them
- We are still flexible enough to [process messages concurrently](https://sbtinstruments.github.io/aiomqtt/subscribing-to-a-topic.html#processing-concurrently) in an individual way
- It's a non-breaking change

The interface looks like this:

```python
router = aiomqtt.Router()


@router.match("humidity/+/inside/#")
async def handle(message, client, *args):  # automatically extract wildcards into *args
    print(message.payload)


async with aiomqtt.Client("test.mosquitto.org", routers=[router]) as client:
        await client.route(message)
```

Where we can process messages concurrently e.g. like this:

```python
router = aiomqtt.Router()


@router.match("humidity/+/inside/#")
async def handle(message, client, *args):  # automatically extract wildcards into *args
    print(message.payload)


async def work(client):
    async for message in client._messages():
        await client.route(message)


async with aiomqtt.Client("test.mosquitto.org", routers=[router]) as client:
    async with asyncio.TaskGroup() as tg:
        tg.create_task(work(client))
        tg.create_task(work(client))
```

(`.messages` is currently broken, for now we use `._messages()` to be able to use it multiple times.)

Glad to hear feedback on this 🙂

### DIFF
--- a/aiomqtt/__init__.py
+++ b/aiomqtt/__init__.py
@@ -8,6 +8,7 @@ from .client import (
 )
 from .exceptions import MqttCodeError, MqttError, MqttReentrantError
 from .message import Message
+from .router import Router
 from .topic import Topic, TopicLike, Wildcard, WildcardLike
 
 # These are placeholders that are managed by poetry-dynamic-versioning
@@ -19,6 +20,7 @@ __all__ = [
     "__version_tuple__",
     "Client",
     "Message",
+    "Router",
     "ProtocolVersion",
     "ProxySettings",
     "TLSParameters",
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -34,6 +34,7 @@ from paho.mqtt.subscribeoptions import SubscribeOptions
 
 from .exceptions import MqttCodeError, MqttConnectError, MqttError, MqttReentrantError
 from .message import Message
+from .router import Router
 from .types import (
     P,
     PayloadType,
@@ -134,6 +135,7 @@ class Client:
         password: The password to authenticate with.
         logger: Custom logger instance.
         identifier: The client identifier. Generated automatically if ``None``.
+        routers: A list of routers to route messages to.
         queue_type: The class to use for the queue. The default is
             ``asyncio.Queue``, which stores messages in FIFO order. For LIFO order,
             you can use ``asyncio.LifoQueue``; For priority order you can subclass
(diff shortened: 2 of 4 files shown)


empicano:
Another possibility would be:

```python
async def handle(message, client, *args):  # automatically extract wildcards into *args
    print(message.payload)


async with aiomqtt.Client(
    "test.mosquitto.org",
    handlers={"humidity/+/inside/#": handle}
) as client:
        await client.route(message)
```

- This could spare users the possible confusion of "Do I use a router or a client, what's the difference?"
- What's nice about the other approach is that the filter (`humidity/+/inside/#`) is written directly next to the handler function, which makes it immediately clear which messages are handled in this function without jumping around in the code. On the other hand, with this option we have all the filters in one place.
- This avoids the problem of defining two routers with overlap in filters

I currently prefer this option.

---

Aside: We should be able to optimize message to handler matching with a trie.

empicano:
Yet another possibility, assigning handlers to subscriptions:

```py
async def handle(message):
    print(message.payload)
    if isinstance(message, Request):
        await message.reply("foo")


    await client.subscribe("humidity/#", handle)
        await client.route(message)
```

We'd have to potentially duplicate messages internally to forward exactly one message per subscription ID. From spec:

> When Clients make subscriptions with Topic Filters that include wildcards, it is possible for a Client’s subscriptions to overlap so that a published message might match multiple filters. In this case the Server MUST deliver the message to the Client respecting the maximum QoS of all the matching subscriptions [MQTT-3.3.4-2]. In addition, the Server MAY deliver further copies of the message, one for each additional matching subscription and respecting the subscription’s QoS in each case.

PR #312, opened 2024-07-01
Fix: Iterating over received messages multiple times does not yield proper results

ryan-summers:
Because of the iterator construction in `__init__`, any attempt to iterate over `client.messages` after the initial iteration will yield no data.

Here's an MVP of the actual defect:
```py
import sys
import os

# Change to the "Selector" event loop if platform is Windows
if sys.platform.lower() == "win32" or os.name.lower() == "nt":
    from asyncio import set_event_loop_policy, WindowsSelectorEventLoopPolicy
    set_event_loop_policy(WindowsSelectorEventLoopPolicy())
import asyncio
import aiomqtt


async def main():
        async def listen():
                print(message.payload)

        listen_task = asyncio.create_task(listen())

        try:
            await asyncio.wait_for(listen_task, timeout=0.5)
        except asyncio.TimeoutError:
            listen_task.cancel()

        # Should continue forever
        await listen()
        raise Exception("Listen returned too early!")


asyncio.run(main())
```

Running the MVP fails with current `main`, but works properly after this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+* Fixed an issue where iterating over received messages would yield nothing after the first
+iteration.
+
 ## [2.1.0] - 2024-04-24
 
 ### Changed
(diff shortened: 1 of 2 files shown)


empicano:
LGTM 👍 Thank you for your first contribution to aiomqtt! 🎉

PR #321, opened 2024-07-17
Update FastAPI doc to use dependency injection

odie5533:
Update the doc example for integration with FastAPI to use FastAPI's dependency injection for the MQTT client

Closes https://github.com/sbtinstruments/aiomqtt/issues/320


empicano:
Perfect, LGTM 👍 Thanks for your contribution to aiomqtt! 🎉

PR #322, opened 2024-08-01
add unix transport hint

spacemanspiff2007:
Update hint from [paho](https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L740)


frederikaalund:
LGTM. Thank you!

PR #339, opened 2024-10-23
Add Proxy Port Parameter

NickJHoffmann:
This allows `proxy_port` to be passed to the underlying paho `proxy_set()` function call according to [the Paho MQTT docs](https://eclipse.dev/paho/files/paho.mqtt.python/html/client.html#paho.mqtt.client.Client.proxy_set). Also addresses https://github.com/empicano/aiomqtt/issues/337


empicano:
Thanks for your contribution to aiomqtt! 🙏

PR #343, opened 2024-11-14
Fix: retained message delayed on TLS connection

wpickard:
Hi, my name is William Pickard and I work for Toro Robotics Colorado.

We noticed that on subscription to topics with a retained message over TLS, the retained message for that topic is received after a considerable delay (seems to be until after the first keepalive).

A review of the `paho-mqtt` project turned up and older PR addressing this issue:

See: https://github.com/eclipse-paho/paho.mqtt.python/issues/131

This PR ports the above fix from `paho-mqtt` to the `aiomqtt` client.

Minimal steps to reproduce:

```python
import aiomqtt
import asyncio
from async_timeout import timeout
import ssl

HOSTNAME = 'test.mosquitto.org'
TOPIC = '/tests/aiomqtt/test_delayed_retain_ssl'

# The TLS cert on test.mosquitto.org seems broken
CONTEXT = ssl.create_default_context()
CONTEXT.check_hostname = False
CONTEXT.verify_mode = ssl.VerifyMode.CERT_NONE

async def test_retained_sub():
    async with aiomqtt.Client(HOSTNAME, 8883, tls_context=CONTEXT) as client:
        await client.publish(TOPIC, payload='hello', retain=True)
        print('Published retained message')
        await client.subscribe(TOPIC)
        
        # This isn't a perfet timing test. It should be almost immediate.
        # The timeout can be removed to observed the total delay.
        async with timeout(1):
                print(f'Received {message.payload.decode()}')
                break
            
asyncio.run(test_retained_sub())
```

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -669,7 +669,14 @@ class Client:
             # usually a sign that the underlying connection broke, therefore we
             # disconnect straight away
             try:
-                client.loop_read()
+                while True:
+                    client.loop_read()
+                    # Client.loop_read() will not always consume all pending
+                    # decoded bytes from SSLSockets
+                    # See: https://github.com/eclipse-paho/paho.mqtt.python/issues/131
+                    if hasattr(client._sock, "pending") and client._sock.pending() > 0:  # type: ignore[union-attr] # noqa: SLF001
+                        continue
+                    break
             except Exception as exc:  # noqa: BLE001
                 if not self._disconnected.done():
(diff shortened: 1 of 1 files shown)


tsturzl:
Experiencing this issue also, this fix seems to resolve the issue.

empicano:
Hi William and Travis, thanks for your contribution, and sorry for the long wait. The fix is included in release v2.4.0 🙂

PR #350, opened 2024-12-09
Migrate poetry to uv

pavel-anchev:


empicano:
I'm terribly sorry for the long wait, lots to do at work and then Christmas. This is a very good start, thank you! 😎

I've added a commit with some small tweaks: Inside `pyproject.toml`, the ruff version was translated from `^0.4.8` to `>=0.4.8,<1.0.0`, but the exact translation is `>=0.4.8,<0.5.0` (same for `sphinx-copybutton`). This incidentally updated ruff "too far" and which raised some new lint errors. In fact, it's a good idea to update the ruff version (and fix these new lint errors), but let's do that in a different PR 🙂

Also (slightly nitpicky): Inside CONTRIBUTING.md we don't need to run `uv python install` and `uv venv`. `uv sync` does that automatically for us. I find that really cool that uv manages the Python version automatically!

---

As far as I can see, this is what's left to do:

1. Automate the version incrementation somehow. With poetry we used [poetry-dynamic-versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) to automatically declare aiomqtt's version based on the GitHub release when we publish a new version. That's why the version is declared as `0.0.0` in the `pyproject.toml` 🙂 Having something similar would be great to avoid duplication and user error
2. Get the test and documentation workflows running
3. Test publishing aiomqtt with workflow to [Test PyPI](https://test.pypi.org/)

You've asked for help (which is great!) in the issue. I know that I've tagged the issue as "good first issue" and these remaining points are not trivial. Do you want to try to tackle them with some pointers from me, or do you want to leave them to me? Both is ok!

empicano:
Please excuse again my delay in reviewing, and thanks for your contribution to aiomqtt! 🎉🙏

PR #357, opened 2025-03-25
Consistently use sock.fileno() to identify socket for monitoring

airtower-luna:
In theory, mixing `sock` and `sock.fileno()` should be fine. In practice pygobject [has a bug](https://gitlab.gnome.org/GNOME/pygobject/-/issues/689) where `loop.add_reader` or `loop.add_writer` fail in that case if using the `GLibEventLoop` in combination with Python 3.13. Tidying up to avoid the bug.

### DIFF
--- a/aiomqtt/client.py
+++ b/aiomqtt/client.py
@@ -686,12 +686,12 @@ class Client:
         # paho-mqtt may call this function from the executor thread on which we've called
         # `self._client.connect()` (see [3]), so we can't do most operations on
         # self._loop directly.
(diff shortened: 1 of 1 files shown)


empicano:
Workflows fail because the CI is a bit out of date, I'll fix those in another MR.

LGTM, thanks for your contributions to aiomqtt! 🙏

Sokjomblo:
hello low profile []()

Sokjomblo (commented):
:))

PR #363, opened 2025-05-28
feature: basic MQTT 5.0 CONNACK properties

### LINKED ISSUE #362: Feature: Return CONNACK properties in `__aenter__`.
I would like to get the MQTT 5.0 properties returned in the CONNACK packet.
My use case is setting payload sizes based on the "Maximum Packet Size" property.

```py
async with Client("broker") as (client, properties):
    pass
```

It should be an easy modification, just pass the properties parameter from `on_connect` to the future and return it in `__aenter__` (I used a tuple but it doesn't matter).

peter-facko:
I tried implementing a minimal version.
Is it fine to do it like this?

closes #362

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -460,9 +460,15 @@ async def test_messages_view_len() -> None:
         # Publish a message and wait for it to arrive
         for index in range(count):
             await client.publish(topic, None, qos=2)
-            await asyncio.wait_for(client.fut, timeout=1)
+            await asyncio.wait_for(client.fut, timeout=2)
             assert len(client.messages) == index + 1
         # Empty the queue
         for _ in range(count):
             await client.messages.__anext__()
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -460,9 +460,15 @@ def _on_message(
         # Publish a message and wait for it to arrive
         for index in range(count):
             await client.publish(topic, None, qos=2)
-            await asyncio.wait_for(client.fut, timeout=1)
+            await asyncio.wait_for(client.fut, timeout=2)
```
peter-facko, line 463: I had this test failing with `timeout=1`, couldn't figure out why.

empicano:
I know this unfortunately got overridden by v3; Thank you anyways for your contribution to aiomqtt!! 🙏

PR #371, opened 2025-08-22
Fix the type hint for Message.payload

### LINKED ISSUE #370: Incorrect type hint for `Message.payload`
Hello, I'm not very familiar with Python, so please correct me if I'm wrong.
Currently, the type hint for `Message.payload` is `PayloadType`.
However, `_from_paho_message()` assigns `paho.mqtt.client.MQTTMessage.payload` directly to `Message.payload`. According to the [paho-mqtt documentation](https://eclipse.dev/paho/files/paho.mqtt.python/html/client.html#paho.mqtt.client.MQTTMessage.payload) , `MQTTMessage.payload` is always of type `bytes`.
Therefore, I believe the correct type hint for `Message.payload` should be `bytes`.

meronepy:
This fixes the type hint for `Message.payload`, which was previously `PayloadType`.
Since `paho.mqtt.client.MQTTMessage.payload` is always `bytes`, the correct type should be `bytes`.

Closes #370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix the type hint for `Message.payload` from `PayloadType` to `bytes` (@meronepy in #370)
+
 ## [2.4.0] - 2024-05-03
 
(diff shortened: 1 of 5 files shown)


empicano:
Looks good to me! Congrats on your first PR, and thanks for your contribution to aiomqtt! 🎉 🙂

PR #376, opened 2025-10-11
Implement v3

empicano:
- [x] Switch protocol library from paho to [mqtt5](https://github.com/empicano/mqtt5)
- [x] Flow control via Receive Maximum
- [x] Reconnection
- [x] Manual acknowledgments and access to PUBACK, PUBREC, PUBREL, PUBCOMP, SUBACK, UNSUBACK, CONNACK
- [x] Publish LWT on exiting context manager with exception
- [x] Only `bytes` payloads (no obscure automatic encoding)
- [x] Rename `Client.messages` to `Client.messages()` to align with `dict.keys()`
- [x] PingReq / PingResp
- [x] Document how to route messages
- [x] Document how to persist with QoS=1 (send) and QoS=2 (send/receive)
- [x] Add migration guide


empicano:
Thanks for checking in, I was hoping to get your opinion on this! :+1:

For v3, I'm aiming for more control and less magic. One of the issues I see with aiomqtt at the moment is that using QoS>0 in a sensible way is actually quite difficult. Currently, we e.g. auto-ack messages on arrival, which undermines the "at least once" and "exactly once" guarantees.

So, that for some background 🙂 Related to publishing, the current draft has the following methods on the client:

```
async def publish(...) -> PubAckPacket | PubRecPacket | None
async def puback(...) -> None
async def pubrec(...) -> None
async def pubrel(...) -> PubCompPacket
async def pubcomp(...) -> None
```

These methods behave like so:

- `publish()` and `pubrel()` methods return once we receive the corresponding acknowledgment
- `puback()`, `pubrec()`, `pubcomp()`, and `publish()` with QoS=0 return once the packet is written to the socket
- `publish()` blocks when we hit the broker's [Receive Maximum](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901083) value
- All methods raise (both in the `reconnect=True` and `reconnect=False` case) when we disconnect

This gives gives full control over retries _and flow control_, with the caveat that the user has to implement retries themselves. Retries with a recreated client need persistence (which aiomqtt/[paho-mqtt don't provide](https://github.com/eclipse-paho/paho.mqtt.python?tab=readme-ov-file#known-limitations)), so I prefer to keep this flexible rather than implementing something that only works if the client is not recreated.

Specifically on your published-obj idea: Currently, `publish()` returns `PubAckPacket | PubRecPacket | None` depending on the QoS level. I believe this is relatively flexible, because it let's us implement something similar to your idea manually on top by scheduling a `publish()` in an [asyncio.Task](https://docs.python.org/3/library/asyncio-task.html#task-object) (which allows us to use the task's `.done()`, `.cancel()`, etc. methods to get information about the status). Does that make sense?

In the current draft, the client is exposing `.connected()` and `.disconnected()` methods, which return once the client connects/disconnects. I'm still a bit hesitant about those, but I don't see a better way to do retries.

I'm still tinkering on all this to get some rough edges out, but that's the current idea 🙂 Please feel free to challenge anything I've said here, I'm very happy to get eyes on this!

---

Side note:

If the user doesn't want to deal with manual acks, then they can tell the broker to downgrade all incoming messages to QoS=0 via the [SUBSCRIBE packet's Maximum QoS property](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901169). IMO, this doesn't loose much to the current state of auto-acking on arrival:

```python
import asyncio
import aiomqtt


async def main():
        await client.subscribe("1234/#", max_qos=aiomqtt.QoS.AT_MOST_ONCE)
        async for message in client.messages():
            print(message.payload)


asyncio.run(main())
```

aschamberger:
> Some thoughts about the two features...
> 
> Maybe add a `reauth(authentication_method: str, authentication_data: bytes)` to update `self._auth*` (for subsequent reconnects) and send the AuthPacket. The logic to detect auth data changes should be outside of aiomqtt.
> 
> Shouldn't just adding an additional argument `ssl: ssl.SSLContext | None` to the constructor, storing to `self._ssl` and passing it in `_connect()` to `asyncio.open_connection()` be sufficient? Handling certificate changes that occur during runtime: add `updateSSLContext(ssl: ssl.SSLContext)` and store to `self._ssl` plus a (re-)connect logic on the `SSLError` exception.

@empicano Should I add an issue for adding reauth() and updateSSLContext()?

empicano:
Good idea to create issues for these, yes! 👍

aschamberger:
Added #394  and #395 to track the feature requests for reauth and SSL cert expiry.

PR #382, opened 2025-12-08
Load TLS certs from the executor thread

### LINKED ISSUE #379: Blocking in event loop when TLS enabled
Version: aiomqtt==2.4.0

This is a minimal code snippet that enables TLS (more below):
```
import asyncio
import aiomqtt

async def main():
    async with aiomqtt.Client(
        hostname="test.mosquitto.org",
        port=8883,
        tls_params=aiomqtt.TLSParameters(),
    ) as client:
            print(message.payload)

asyncio.run(main())
```
The above will end up:
- Calling the `aiomqtt.Client` constructor
- That function calls `tls_set` on the underlying mqtt client
- The `tls_set` fucntion performs blocking I/O  calling `context.load_default_certs`

However, actually ran into this in Home Assistant. Home Assistant has code to catch blocking in the event loop which shows the problematic path that is blocking from async code:
```
2025-10-28 21:13:36.582 WARNING (MainThread) [homeassistant.util.loop] Detected blocking call to load_default_certs with args ( For developers, please see https://developers.home-assistant.io/docs/asyncio_blocking_operations/#load_default_certs Traceback (most recent call last): File " File " File "/usr/src/homeassistant/homeassistant/main.py", line 229, in sys.exit(main()) File "/usr/src/homeassistant/homeassistant/main.py", line 215, in main exit_code = runner.run(runtime_conf) File "/usr/src/homeassistant/homeassistant/runner.py", line 271, in run return loop.run_until_complete(setup_and_run_hass(runtime_config)) File "/usr/local/lib/python3.13/asyncio/base_events.py", line 712, in run_until_complete self.run_forever() File "/usr/local/lib/python3.13/asyncio/base_events.py", line 683, in run_forever self._run_once() File "/usr/local/lib/python3.13/asyncio/base_events.py", line 2050, in _run_once handle._run() File "/usr/local/lib/python3.13/asyncio/events.py", line 89, in _run self._context.run(self._callback, *self._args) File "/usr/local/lib/python3.13/site-packages/roborock/mqtt/roborock_session.py", line 105, in _run_task async with self._mqtt_client(self._params) as client: File "/usr/local/lib/python3.13/contextlib.py", line 214, in aenter return await anext(self.gen) File "/usr/local/lib/python3.13/site-packages/roborock/mqtt/roborock_session.py", line 155, in _mqtt_client async with aiomqtt.Client( File "/usr/local/lib/python3.13/site-packages/aiomqtt/client.py", line 323, in init self._client.tls_set( File "/usr/local/lib/python3.13/site-packages/paho/mqtt/client.py", line 1295, in tls_set context.load_default_certs()
```

I don't know of an easy change here. A common pattern is for the caller to just workaround this by calling the function from `run_in_executor` however for this case:
- The Client constructor is sync (fine)
- however, it expects to be called from inside the asyncio event loop 

Perhaps the call to `tls_set` can be moved somewhere else and can load that from the executor thread.

allenporter:
This will avoid doing blocking I/O in the event loop when connecting to the client. This was previously set in the constructor, and is moved to run in an executor thread at the same time as connect() happens.

An alternative would be to only call `tls_set` but it didn't necessarily seem worth tracking state to do that. (could be done by clearing `_tls_params` though for example)

Fixes #379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - Fix the type hint for `Message.payload` from `PayloadType` to `bytes` (@meronepy in #370)
+- Load TLS certs from the executor thread (@allenporter in #382)
 
 ## [2.4.0] - 2024-05-03
 


ohadbenita:
thanks for working on this 👍 
Is there any ETA for it being released?

allenporter:
Hi, we need to have the project owner review and release the package. Keep in mind this is likely a volunteer effort on their part.

My impression is the impact here is low or none and results in initial logging in home assistant without impact on functionality.

allenporter:
Gentle nudge, post holidays.

